PR #9, opened 2024-01-15
Update README.md for installation instruction

ViralBShah:


ViralBShah:
The docs also need an update to use the github url instead of the MIT one.

askprash:
Thanks @ViralBShah, updated docs as well. There are going to be a couple of code updates coming soon that will come with some doc updates too.

PR #19, opened 2024-02-09
woper.jl : purge NPSS, other updates

aditeyashukla:
- Removed NPSS
- Changed function to take ac and specific off design mission to run 
- option to save off design mission returned values
- initeng changed to bool
- cleaned up code a bit

### DIFF
--- a/src/mission/woper.jl
+++ b/src/mission/woper.jl
@@ -22,8 +22,8 @@ function woper(ac, mi = 1; itermax = 35, initeng = true, saveOffDesign = false)
     parm = ac.parm[:,mi:mi]
     para = ac.para[:,:,mi:mi]
     pare = ac.pare[:,:,mi:mi]
-    parad = ac.para[:,:,1:1]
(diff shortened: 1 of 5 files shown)


src/mission/woper.jl
```diff
@@ -1,41 +1,37 @@
...
-    # NPSS_TS = Base.Process
-    # NPSS_Fan = Base.Process
-    # NPSS_AftFan = Base.Process
-    # NPSS_PT = true
+    pari = ac.pari
+    parg = ac.parg
+    parm = ac.parm[:,mi:mi]
+    para = ac.para[:,:,mi:mi]
+    pare = ac.pare[:,:,mi:mi]
+    parad = ac.para[:,:,1:1]
```
askprash, line 25: This will create copies of the array, making it slower. Could use the new `parad = ac.parad` which directly gives a view of the array. Similarly `ac.pared; ac.parmd` etc should work.

PR #21, opened 2024-03-18
Choosing materials from a material database

askprash:
Chipping away at a more modular and cleaner code. This PR adds the ability to choose from some standard aerospace materials when designing an aircraft in TASOPT. 

Users can add materials of their choice to the `MaterialProperties.toml` database file. 

When creating components in TASOPT checks are performed if the properties provided in the database are sufficient to create a component that uses the material specified in that fashion. E.g., creating a component of type `StructuralAlloy` will only work if properties such as the density, elastic modulus, yield strength etc are provided. Similarly component of type `Conductor` can only be created if the database contains properties about the material's resistivity.

### DIFF
--- a/src/misc/materials.jl
+++ b/src/misc/materials.jl
@@ -1,6 +1,9 @@
 abstract type AbstractMaterials end
 
-MaterialProperties = TOML.parsefile("src/material_data/MaterialProperties.toml")
+using TOML, DocStringExtensions
+
+__abs_path_prefix__ = dirname(@__DIR__)
+MaterialProperties = TOML.parsefile(joinpath(__abs_path_prefix__,"material_data/MaterialProperties.toml"))
 
 """
 $TYPEDEF
--- a/src/misc/materials.jl
+++ b/src/misc/materials.jl
@@ -1,5 +1,3 @@
-abstract type AbstractMaterials end
-
 using TOML, DocStringExtensions
 
 __abs_path_prefix__ = dirname(@__DIR__)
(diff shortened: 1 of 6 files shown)


src/misc/materials.jl
```diff
@@ -0,0 +1,201 @@
+abstract type AbstractMaterials end
```
ian-ross, line 1: I don't think this is giving you anything. Is there any circumstance under which you would want to use a generic "material" without knowing anything about it in more detail? There is no real relationship between the different concrete material types you've defined, so unless you have something in mind for this generic case, I don't think you need to derive them from a common abstract type like this.

src/misc/materials.jl
```diff
@@ -0,0 +1,201 @@
...
+MaterialProperties = TOML.parsefile(joinpath(__abs_path_prefix__,"material_data/MaterialProperties.toml"))
+
+"""
+$TYPEDEF
+
+Generic structural alloy.
+
+$TYPEDFIELDS
+"""
+struct StructuralAlloy <: AbstractMaterials
```
ian-ross, line 15: I don't know if I would have done things in exactly this way, but I think this is good. It makes the individual materials into simple concrete structures. My solution would probably have been a bit more airy-fairy and abstract, with some sort of separate `MaterialProperties` structure for each material, but I like your way better.

ian-ross (approved):
Yeah, I think this looks good. It's a nice small bite of the "making things more structured" pie. A good start!

src/misc/materials.jl
askprash: Agreed, I don't really have a use case for this now. Will get rid of it.

PR #23, opened 2024-03-22
Improved fuel storage options for cryogenic fuels

ngomezve:
This PR revives and revamps fuel storage in a double-walled tank with a vacuum layer. In addition, these changes have been introduced:
- The user can choose whether to size a cryogenic fuel tank for a given boiloff rate or just use the provided insulation thicknesses.
- The tests for the fuselage tanks have been expanded.
- Fixed bug in wsize where xconend, xapu and xeng local variables were not updated when the fuselage was resized.


askprash:
These look like some really awesome additions! 🎉 Thanks @ngomezve!

Looking at all the changes though, these really should to be separate PRs. May have made sense for each one of the bullets below to be their own bite-sized PR. [This article](https://www.atlassian.com/blog/git/written-unwritten-guide-pull-requests) may be worth perusing!

> This PR revives and revamps fuel storage in a double-walled tank with a vacuum layer. In addition, these changes have been introduced:
> 
> * The user can choose whether to size a cryogenic fuel tank for a given boiloff rate or just use the provided insulation thicknesses.
> * The cabin and fuselage length can be optionally sized for the maximum number of passengers.
> * The cabin can have two decks.
> * The user no longer has to specify fuel temperature and density with cryogenic fuels. These are calculated in read_input().
> * The tests for the fuselage tanks have been expanded.

src/structures/tankWmech.jl
```diff
@@ -140,11 +144,278 @@ function tankWmech(gee::Float64, ρfuel::Float64,
...
+      opt = Opt(:LN_NELDERMEAD, length(initial_x))
+      opt.lower_bounds = lower
+      opt.upper_bounds = upper
+      opt.ftol_rel = 1e-9
+      opt.maxeval = 100  # Set the maximum number of function evaluations
+
+      opt.min_objective = obj
+
+      (minf,xopt,ret) = NLopt.optimize(opt, initial_x)
+      Ninterm = xopt[1]
```
askprash, line 417: This will be a `float`! Shouldn't we take the `ceil`?
```suggestion
      Ninterm = ceil(Int, xopt[1])
```

ngomezve: I decided to keep it a Float64 instead of ceiling it to an integer. The reasoning was that making it an integer would introduce discontinuities and I was concerned it would affect the sizing loop convergence.

PR #26, opened 2024-03-29
Double decker cabin + Test improvements

ngomezve:
- Added an optional input to make the cabin have two decks. In the future, the two decks could be sized independently but currently they are the same size.
- Removed equality requirement in some tests.
- Fixed IO test, wherein a csv file would be created and not removed. Also updated number of fields.


src/structures/update_fuse.jl
```diff
@@ -74,7 +74,15 @@ function update_fuse_for_pax!(pari, parg, parm, fuse_tank)
...
     aisle_halfwidth = parg[igaislehalfwidth]
 
-    maxpax = parg[igWpaymax]/parm[imWperpax,1] #maximum number of passengers
+    if pari[iidoubledeck] == 1 #if the aircraft is a double decker
+        #passenger count to size cabin is half of the maximum
+        paxsize = ceil(parg[igWpaymax]/parm[imWperpax,1] / 2) 
+    else
+        paxsize = parg[igWpaymax]/parm[imWperpax,1] #maximum number of passengers
+    end
+    #TODO this double deck model assumes that both decks have a width equal to the fuselage diameter;
```
askprash, line 83: Yes, I'll raise an issue so we remember to do this at some point in the future.

askprash (approved):

PR #29, opened 2024-04-02
Improved models for cryogenic fuel boiloff

ngomezve:
This PR applies some minor improvements to the models for cryogenic fuels:
- Boiloff fuel is only vented when it would result in an increase in tank pressure (if the volume of the boiloff gas at saturation is greater than the volume that is made available by liquid boiloff and fuel burn).
- The mass of the ullage gas inside the tank is now accounted for.
- The latent heat of vaporization of hydrogen and methane is now a function of tank pressure.
- Added Mylar as a material option for the insulation, specifically for vapor barriers.
- Font size in tank diagrams for docs has been increased.


ngomezve:
I converted the PR back to draft to increase test coverage based on Codecov report.

PR #31, opened 2024-04-04
Fix LTO engine calculations

askprash:
Engine LTO calculations were broken since the removal of NPSS. `LTO.jl` has now been fixed to use `tfcalc` instead.

Also added a version variable in `TASOPT.jl` and fixed a misspelled "dictionary"

### DIFF
--- a/src/mission/LTO.jl
+++ b/src/mission/LTO.jl
@@ -100,7 +100,7 @@ function EINOx4(P3_kPa, T3_K, sp_humidity = 0.00634)
 
     H = -19.0*(sp_humidity - 0.00634)
 
-    EI_NOx = exp(H)*P3_kPa^0.4 * (a*T3_K^4 + b*T3_K^4 + c*T3_K^2 + d*T3_K + e)
+    EI_NOx = exp(H)*P3_kPa^0.4 * (a*T3_K^4 + b*T3_K^3 + c*T3_K^2 + d*T3_K + e)
     return EI_NOx
 end  # function EINOx4
 
--- a/src/mission/LTO.jl
+++ b/src/mission/LTO.jl
@@ -71,6 +71,8 @@ end
 
 Returns the EI(NOₓ) for a CFM56 level engine.
 Uses a third order polynomial in T₃ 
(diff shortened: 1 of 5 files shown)


src/mission/LTO.jl
```diff
@@ -1,45 +1,111 @@
...
+function EINOx4(P3_kPa, T3_K, sp_humidity = 0.00634)
+    a = 4.85354237e-11
+    b = -6.51089333e-08
+    c =  7.19366066e-06
+    d = 2.06850617e-02
+    e =  -6.69412110e+00
+
+    H = -19.0*(sp_humidity - 0.00634)
+
+    EI_NOx = exp(H)*P3_kPa^0.4 * (a*T3_K^4 + b*T3_K^4 + c*T3_K^2 + d*T3_K + e)
```
sareider, line 103: T3_K after b should be ^3

src/mission/LTO.jl
```diff
@@ -1,45 +1,111 @@
...
+    return EI_NOx
+end  # function EINOx4
+
+function EINOx3(ac::aircraft, ip::Int, sp_humidity = 0.00634)
+       P3_kPa = ac.pared[iept3, ip]/1000.0
+       T3_K   = ac.pared[ieTt3, ip]
+       EINOx3(P3_kPa, T3_K, sp_humidity)
```
sareider, line 110: Need another function like this for EINOx4

src/mission/LTO.jl
askprash: Good catch!

askprash, line 103:
```suggestion
    EI_NOx = exp(H)*P3_kPa^0.4 * (a*T3_K^4 + b*T3_K^3 + c*T3_K^2 + d*T3_K + e)
```

src/mission/LTO.jl
askprash: Added!

askprash:
- [ ] Add a temperature range check that throws a warning if $T_3 \notin T_{3,limits}$

PR #40, opened 2024-04-22
Homogeneous model of cryogenic tanks

ngomezve:
This PR introduces a homogeneous model for a cryogenic tank. This treats the fluid inside the tank as a mixture of a saturated liquid and a saturated vapor. Through this, TASOPT can model the evolution of pressure in time inside the tank, as well as calculate the required venting more accurately. In addition, this PR also:
- Introduces a new CryoTank module, which is where the homogeneous model is stored.
- Changes the way that venting is treated in mission; now venting is only added at the end of the flight profile. Note that this results in a slightly inaccurate profile.
- Adds a natural convection model to the tank thermal models, needed for when the aircraft is stationary on the ground.
- Moves the calculation of the tank fuel densities to wsize() from read_input().
- Eliminates old functions in the src/fuel folder that have now been replaced.
- Introduces tests for the new CryoTank module.
- Adds DifferentialEquation.jl package for ODE solving. The ODE solver in the PEMFC models has been replaced by this.
- Documents the new tank model.


ngomezve:
My next PR will introduce a regression test for a hydrogen aircraft sizing that will improve test coverage.

ngomezve:
The ODE solver in DifferentialEquations.jl is now used instead of the custom RK4 one.

src/cryo_tank/pressure.jl
```diff
@@ -0,0 +1,232 @@
...
+    """Quality of mixture being vented"""
+    xvent::Float64
+    """Pressure fudge factor"""
+    α::Float64
+end
+
+"""
+Structure with inputs for tank pressure calculation.
+"""
+struct pressure_inputs
```
askprash, line 22: Can we have a better name for this?

PR #42, opened 2024-04-30
Modified atmosphere model for different sea-level temperatures

ngomezve:
This PR modifies the atmosphere model in TASOPT to model different sea-level temperatures. In the model, the pressure distribution in the atmosphere from standard sea-level conditions is maintained and the sea level temperature is modified to match the desired temperature at the takeoff altitude. The density, speed of sound, and viscosity of air are modified accordingly.

The new atmosphere model replaces the old corrections in mission() for different takeoff temperatures. It is integrated into the hydrogen tank models, which may be designed for a different SL temperature than the rest of the airframe. 

The default, regional and widebody TOML files have been modified (TO temperature changed from 288 to 288.2 K) so that the new and old models give the same results. The test files have been updated accordingly.


ngomezve:
I rewrote the code so that the atmos input is DeltaT instead of TSL

Issue #33, opened 2024-04-12
IO docs and examples


argonaut22:
Documenting the conveniences tosave/load models, and get data out of TASOPT en masse would lower the bar for usage. Placeholder pages exist.

- [x] short writeup and examples for `output_csv`
- [x] " for quickloading/saving
- [x] " for readable .toml loading/saving
- [x]  note about `index.inc` and convenient loading a la `include(__TASOPTindices__)`

PR #44, opened 2024-05-02
IO documentation added

argonaut22:
addresses #33 

(this time, without duplicate commits)

### DIFF
--- a/docs/src/data_io/data_basics.md
+++ b/docs/src/data_io/data_basics.md
@@ -12,6 +12,46 @@ All data, both prescribed inputs *and* computed outputs,  are stored in 5 arrays
   - **`para`**`::AbstractArray{Float64, 3}`:  **aerodynamic** performance quantities. The second dimension captures the variation over a mission. The third dimension allows the specification of multiple mission profiles.
   - **`pare`**`::AbstractArray{Float64, 3}`:  **engine** perfomance quantities. As for `para`, the second and third dimensions capture flight-point and mission dependencies, respectively.
 
+Data in the `par` arrays are accessed via Integer indices defined at `src/misc/index.inc`. These indices can be added to a namespace via an `include` statement:
+
+```julia
+using TASOPT
+#using __TASOPTroot__, which fetches the src directory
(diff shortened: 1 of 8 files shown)


docs/src/data_io/data_basics.md
```diff
@@ -0,0 +1,66 @@
...
+#using __TASOPTroot__, which fetches the src directory
+include(joinpath(__TASOPTroot__, "misc/index.inc"))
+
+#or more concisely
+include(__TASOPTindices__)
+```
+
+The variable names of these indices indicate which `par` array they should access and hint at the quantity in question. For example, `ieTfuel` evaluates to `2` and retrieves the model's fuel temperature via `pare[ieTfuel]`. 
+
+Note that for the multi-dimensional `par` arrays, indexing with a single Integer only retrieves the value for the first flight point of the first mission (namely, the design mission). Add'l indexing is required to access data from different flight points or missions. Indices for specific flight points are defined in `index.inc` and should be used when indexing `pare` or `para`, e.g., `ipstatic` for static ground condition or `ipcruise1` for the start of cruise.
```
askprash, line 28:
```suggestion
Note that for the multi-dimensional `par` arrays, indexing with a single Integer only retrieves the value for the first flight point of the first mission (namely, the design mission). Additional indexing is required to access data from different flight points or missions. Indices for specific flight points are defined in `index.inc` and should be used when indexing `pare` or `para`, e.g., `ipstatic` for static ground condition or `ipcruise1` for the start of cruise.
```

docs/src/data_io/data_io.md
```diff
@@ -1,36 +1,96 @@
...
+#load and size the default aircraft model
+using TASOPT
+ac2 = load_default_model()
+size_aircraft!(ac2)
+
+#quicksave aircraft at 4.55pm on a Friday
+filepath2 = "/path/to/your/new/quicksave.toml"
+quicksave_aircraft(ac2, filepath2)
+
+# - w e e k s - p a s s -
```
askprash, line 38: Is this actually very slow? Any idea why?

PR #55, opened 2024-05-28
Replacing enthalpy of combustion in PFEI by LHV

ngomezve:
This PR makes a change to the way PFEI is calculated. Rather than using the enthalpy of combustion of the fuel, which depends on the fuel initial temperature, the PFEI is now calculated with the lower heating value, which is a constant for a given fuel type.


askprash:
Looks good! Were you going to add anything else or can we mark this ready to pull instead of draft?

PR #56, opened 2024-05-31
Cryogenic fuels: hold times, more insulation options and tests

ngomezve:
This PR addresses the models for aircraft with cryogenic fuels in insulated tanks:
- The effect of hold times at departure or arrival can now be modeled; this will affect fuel venting.
- Radiation to the fuselage has been eliminated from the thermal models; the effect was negligible and the model was sketchy.
- Added vacuum microspheres as an insulation option.
- Fixed a bug in the balance-field calculation with heat exchangers, wherein the effect of the heat exchanger was not being accounted for.
- Fixed a small bug in the tank time integration, which could ignore the takeoff time.
- Minor changes to gasfun.jl and the models in CryoTank to eliminate type promotion and improve speed.
- New test for hydrogen aircraft sizing.


src/cryo_tank/tanksize.jl
```diff
@@ -74,7 +75,9 @@ function tanksize!(fuse_tank, z::Float64, Mair::Float64, xftank::Float64,
         Winner_tot, lcyl1, tskin, Rinnertank, Vfuel, Winnertank, Wfuel_tot, Winsul_sum, t_head, Whead, Wcyl, Wstiff, Winsul,
         Sinternal, Shead_insul, l_inner = size_inner_tank(fuse_tank, fuse_tank.t_insul)
 
-        if ("vacuum" in fuse_tank.material_insul) || ("Vacuum" in fuse_tank.material_insul) #If tank is double-walled
+        flag_vacuum = check_vacuum(fuse_tank.material_insul) #check if there is a vacuum layer
```
askprash, line 78: This is interesting, I wonder if insulation materials can be something that extends the `Materials` type in a future refactoring? E.g., something like `ThermalInsulators` like `StructuralAlloys` or something. This is fine for now though!

PR #57, opened 2024-06-04
Refactoring the fuselage sizing

askprash:
@aditeyashukla starting a draft PR here so that others can also see what we've been up to and add any suggestions.

This PR refactors the fuselage sizing code to be more modular, easy to read and more efficient.

Some of the updates/features:
- Introduce the `Fuselage`, `FuselageLayout`, and a couple other types that make it easier to define and size the fuselage
- `StructuralMembers` are a generic type that can be used to represent various structural components - e.g., in the fuselage this may be the skin, the floor, etc.
- We introduce `loads.jl` that has lets us define `Weights` and more semantically calculate the moments of weights. Also have some basic coordinate system introduced (e.g., now we can associate loads/ weights with a `Frame` which defaults to a privileged `WORLD` frame. This can allow us to translate between frames more easily. We may potentially replace this with homogenous coordinates if we start to find it useful enough.


askprash:
Fixed #68

aditeyashukla (approved):

Issue #60, opened 2024-06-07
Maximum payload and maximum number of passengers are coupled


ngomezve:
The maximum aircraft payload is currently determined based on the maximum number of passengers, with no option to specify an additional cargo payload. In cases where the cabin in resized, it may be appropriate to distinguish between the exit limit number of passengers of the aircraft and the cargo weight.

PR #63, opened 2024-06-10
Exit limit as an input

ngomezve:
Addresses issue #60.

 A new input `exit_limit` is required that defines the aircraft's exit limit and therefore the maximum number of passengers that can sit in the cabin. The maximum payload is still determined through `max_pax`, but a comment has been added to explain that this is an *equivalent* number of passengers.


ngomezve:
Closing as long outdated.

PR #64, opened 2024-06-13
Double-bubble cryogenic tanks

ngomezve:
This PR removes inconsistencies in the geometry of the cryogenic tanks in the fuselage. Some models assumed that the geometry was a double bubble while others only applied to circular sections. Now, the tank cross-section can be a double bubble in general, following the fuselage section.

This PR also introduces a new function `double_bubble_geom()` that can be used to calculate the geometric properties of a double bubble shape. This function could also be used in `fuseW` for the fuselage design.


ngomezve:
Closing as superseded by PR #80.

PR #67, opened 2024-07-02
Improved cabin layout and double decker models

ngomezve:
This PR introduces better design options if the fuselage layout has to be redesigned; for example, following a change in fuselage radius. The following changes are included:
- The seat height is included as an input parameter to ensure that seats can fit in the cabins.
- If the fuselage bubble has a downward shift, the floor is assumed to lie at the intersection of the two bubbles.
- If the fuselage has two decks, the decks are no longer assumed to be symmetric or have the same number of passengers. Instead, the number of passengers in the lower deck is optimized, along with the lower deck position if the cross section is circular. This optimization is subjected to constraints on cargo bay height and cabin height.
- A new function has been included to find the minimum radius required to have a given number of seats per row. This is useful if the radius has to be redesigned.
- The functions to size the cabin have been moved from IO to structures.


ngomezve:
This PR is now largely outdated due to the fuselage data structure revamping. I made a new PR #78 to revive the H2 models and introduce the features in this PR.

PR #77, opened 2024-08-21
Reduced Input TASOPT

aditeyashukla:
- Changed how read input works
- First reads key in Aircraft Description: `aircraft_type`, then replaces the "default" model for that in case any key missing
- If `aircraft_type` not specified then decides default model based on range:

Regional: range <= 2600 nmi
Narrow Body: range <= 3115 nmi
Wide Body: range <= 8500 nmi

Other changes:
- Added error if design mission pax > maxpax
- Added wide and regional default models to IO folder

### DIFF
--- a/src/IO/read_input.jl
+++ b/src/IO/read_input.jl
@@ -28,11 +28,11 @@ function read_input(k::String, dict::AbstractDict=data,
 end
 
 function get_default_input_file(input)
-    if input == 0 || input == "Regional Aircraft"
+    if input == 0 || lowercase(input) == "regional aircraft"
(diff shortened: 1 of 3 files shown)


aditeyashukla:
@askprash pls lmk if there's a different/better way to do this

src/IO/read_input.jl
```diff
@@ -27,6 +27,20 @@ function read_input(k::String, dict::AbstractDict=data,
     end
 end
 
+function get_default_input_file(input)
+    if input == 0 || input == "Regional Aircraft"
```
askprash, line 31:
```suggestion
    if input == 0 || lowercase(input) == "regional aircraft"
```

PR #80, opened 2024-09-06
Double bubble cryo tanks

ngomezve:
This PR is a revision of PR #64. It removes inconsistencies in the geometry of the cryogenic tanks in the fuselage. Some models assumed that the geometry was a double bubble while others only applied to circular sections. Now, the tank cross-section can be a double bubble in general, following the fuselage section.

This PR also introduces a function that can scale down a fuselage cross section. It makes use of the new ``Fuselage`` data type.


ngomezve:
I removed my double bubble geometry function from utils and replaced it with a scaled geometry function in layout, so as to use the new fuselage data structures

PR #82, opened 2024-09-10
Better stickfig aircraft plots

ngomezve:
This PR adds new options to the `stickfig` function that makes stick figure plots of aircraft. It combines it with the alternative function `high_res_airplane_plot`, that produced a very similar-looking plot and had most of the code duplicated. 
These functionality has been implemented:
-The engine is shown again; it was not being rendered due to a data type issue.
-The span, text, and length annotations are now optional; these can be disabled by changing the optional function inputs.
-The ICAO and FAA airport codes are now found automatically for the design maximum span. It fixes a mistake in the previous code, in which ICAO code D was incorrectly compared to FAA group III; it should be ICAO Code C. The plotting area is changed accordingly.
-The text annotations are now shown to the right of the airplane plot, rather than underneath it, which covered most of the text.
-A grid can be optionally added.

### DIFF
--- a/src/IO/outputs.jl
+++ b/src/IO/outputs.jl
@@ -186,7 +186,9 @@ end
 like [`plot_details`](@ref) to create summary plots to track progress of optimization
 or present results.
 """
-function stickfig(ac::aircraft; ax = nothing, label_fs = 16)
+function stickfig(ac::aircraft; ax = nothing, label_fs = 16, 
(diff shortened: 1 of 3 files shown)


ngomezve:
This is an example of a stick figure with the code currently in `main`:
Example

This is the same example with the changes in the PR:
New

PR #83, opened 2024-09-12
Created thermal insulation materials

ngomezve:
This PR introduces a new type of material object: `ThermalInsulator`. This object contains two fields: density and a `Vector` for a polynomial defining thermal conductivity as a function of temperature. These new materials are used mostly in the fuselage tank models, as insulators. They replace a set of string that was passed into ad-hoc functions with the material properties. The new materials could easily be used in other parts of the aircraft, such as for cabin thermal insulation.

In addition, the following changes are made:
-The fuselage tank struct has been put into its own separate file and is better commented.


aditeyashukla (approved):
Looks good!

PR #84, opened 2024-09-30
Electric Machines 

askprash:
This PR adds models for electric machines, including motors, generators, inverters and cables. In general, sizing and off-design operation functions are created for each of these components.


docs/src/propulsion/ElectricMachines.md
```diff
@@ -0,0 +1,154 @@
+# Electric Machines
+
+## Permanent-magnet synchronous machines
+
+Permanent-magnet synchronous machines (PMSMs), particularly motors and generators, can be modeled tp estimate their performance and weight. The function [`size_PMSM!()`](@ref propsys.ElectricMachine.size_PMSM!) can be used to find the thickness and lengths of the PMSM components, their masses, as well as to calculate the phase electrical resistance. Once a PMSM has been sized, its off-design performance can be computed using [`operate_PMSM!()`](@ref propsys.ElectricMachine.operate_PMSM!), which calculates power losses tp compute the input power (motor) or output power (generator).
```
askprash, line 5:
```suggestion
Permanent-magnet synchronous machines (PMSMs), particularly motors and generators, can be modeled to estimate their performance and weight. The function [`size_PMSM!()`](@ref propsys.ElectricMachine.size_PMSM!) can be used to find the thickness and lengths of the PMSM components, their masses, as well as to calculate the phase electrical resistance. Once a PMSM has been sized, its off-design performance can be computed using [`operate_PMSM!()`](@ref propsys.ElectricMachine.operate_PMSM!), which calculates power losses and computes the input electrical power required (motor) or output shaft power (generator).
```

src/propsys/PMSM.jl
```diff
@@ -0,0 +1,616 @@
...
+
+$TYPEDFIELDS
+
+Structure that defines a permanent magnet synchronous motor (PMSM).
+"""
+@kwdef mutable struct Motor <: AbstractElectricMachine
+    base::PMSMBase = PMSMBase()
+
+    """Number of multi-phase inverters [-]"""
+    N_inverters::Int = 1
```
askprash, line 223: Are there setups that need more than 1 inverter for a single machine? This motor structure seems unnecessary.

src/propsys/PMSM.jl
```diff
@@ -0,0 +1,616 @@
...
+        return obj.N_slots * obj.A_slot / cross_sectional_area(obj.teeth.Ri + obj.teeth.thickness, obj.teeth.Ri) 
+    elseif sym === :B_gap #Air gap flux density
+        return airgap_flux(obj.magnet.M, obj.magnet.thickness, obj.airgap_thickness)
+    else
+        return getfield(obj, sym)
+    end
+end  # function Base.getproperty
+
+#List of computed fields that are not stored in the structure but are derived from other fields
+const BASE_COMPUTED_FIELDS = [:f, :torque, :N_slots, :N_slots_per_phase, :N_energized_slots, :λ, :B_gap]
```
askprash, line 210: Declaring something as "BASE" to refer to some very specific variables just for convenience is going to be very confusing down the line! I see what you are trying to do here but I'm not sure this setup of having a "base" implementation is really needed just to define the Motor and Generator as two types of PMSM. This introduces a bunch of convenience functions that I think are just going to make it harder to read later.

The motor and generator could be implemented simply via a parametric type added to the PMSM sturct and then the sizing functions will know exactly what to do - what are inputs and what are outputs.

askprash (commented):
Looks pretty good overall! Thanks for completing this. 

Main thing to clean up is the motor/generator implementation as instances of base. The current approach introduces a bunch of code that will make maintenance harder in the future. I have some ideas of how to clean this up using a parametric type in PMS M and then the downstream functions should be straight forward. I'll try to add that in as suggestions in a while.

src/propsys/PMSM.jl
ngomezve: My understanding is that by using multiple inverters, you can reduce the required voltage as the power is divided among them. For example, the GTL group at MIT used thirty single-phase inverters.

src/propsys/PMSM.jl
ngomezve: Have a look at the parametric type implementation in commit [25d7411](https://github.com/MIT-LAE/TASOPT.jl/pull/84/commits/25d7411455e7e8040a1348005514b1bfe8a24795). I agree it is a lot neater!

PR #85, opened 2024-10-03
Sensitivity module

aditeyashukla:
added functions to calculate sensitivity using finite difference (central, relative) as well as plot it
Added markdown file with sample script


docs/src/assets/sensitivity_plot.png
ngomezve, line 1: Replace .png with .svg

ngomezve (changes requested):
Good PR! I suggest you change the png file to an svg so as to better track changes (svg is a vector format)

docs/src/assets/sensitivity_plot.png
aditeyashukla: Thanks! Pushed the change

PR #87, opened 2024-10-09
Generating LTO and BADA style PTF files 

aditeyashukla:
Note: Since we are using the P3T3 method for getting NOx values, it actually just makes sense for a CFM56-7B (since the equation is matched to that)

ALSO:  Added aircraft type in model files


askprash:
@aditeyashukla can we add some tests for the functions here?

PR #88, opened 2024-10-09
Update CI.yml

askprash:
Testing if ditching the julia test action is better


askprash:
not needed anymore given #93

PR #90, opened 2024-10-10
Removing nested loop in mission!

ngomezve:
This PR achieves significant code speedup by eliminating a redundant iteration loop inside `mission!()`. The iteration loop is used to calculate a consistent flight speed and climb angle (by calling `tfcalc!()`) at every sizing loop iteration. However, thrust is very weakly affected by the climb angle (through its cosine, which is close to 1), so instead, it is better to use the climb angle from the previous sizing iteration and let the solution become consistent when the entire aircraft converges. The convergence criterion for the inner loop was also very strict (1e-12) and required at least two iterations for every climb point.

Compared to `main` after the gas calculation speedup, this PR results in up to 2x speed improvement. The median run times on my Intel i7 are shown below for different cases:
|          | main       | dev_climb  |
|----------|------------|------------|
| Default  | 479.095 ms | 225.615 ms |
| Regional | 510.950 ms | 225.712 ms |
| Wide     | 592.615 ms | 291.806 ms |
| Hydrogen | 1.689 s    | 1.194 s    |


aditeyashukla:
I have never been more excited for a PR to merge

PR #91, opened 2024-10-11
Added new benchmarks

ngomezve:
This PR revives the old benchmark functions for the fuselage and drag calculations (now in `benchmark_elements.jl`). It also introduces functions to benchmark the aircraft sizing for all four example models (default, regional, widebody and cryo).


askprash:
@ngomezve  this is great! Could you add a short nod to this in the docs - I'm thinking here: https://mit-lae.github.io/TASOPT.jl/dev/misc/fordevs/ ? Something like:
```
!!! tip "Benchmarks"
    - Some example benchmarking files are provided in `test/benchmark_elements.jl`.
    - Developers can similarly create other benchmarks for new features, models, and functions.
```

PR #92, opened 2024-10-15
Compressor Map Table Interpolation v1

WyattGiroux:
Added the `NcTblMap` and `ecTblMap` functions for table interpolation of speed and efficiency (+ derivatives). Also added the `bilinearBounded.jl` file to perform the bilinear interpolation and the `tftbl.jl` file to generate the const map data structs.

Ran unit tests in test/runtests.jl and got all either passing or broken:

image (3)


ngomezve:
@WyattGiroux This looks promising! Are you planning to implement the new compressor maps into `tfsize()` and `tfoper()`, the turbofan functions? It looks like right now they are still using the old maps, which explains why all tests are passed (the new functions are not being run).

WyattGiroux:
@ngomezve, I am planning on integrating them with oper and size, but need to make some unit tests for the new functions before that. 

In general, I've now swapped over to the NaturalNeighbours.jl library to do direct interpolation in julia. Now instead of using the bilinear functions I made, the TblMap functions are now calling interpolator functions from this package. The timing is slower (80 microseconds) but the maps are being very closely matched (max[RMSE] = 0.01-ish) using very few gridding sample points (~50^2). The package precompile time is around 1 minute.

Unit tests to come next, but good progress is being made

WyattGiroux:
We're now faster than the original drela method and have hashing implemented such that the gridded map files only have to generate for 30 minutes once. Derivatives are a bit noisy; will try various samplings and gridding fineness later tonight. If all goes well, I'll add docstrings, unit tests, and TASOPT implementation tonight.

image (5)

WyattGiroux:
All unit tests are in and functionality is verified against NPSS data to be reasonable (within 2% relative error). Implementing into tfoper, etc. next. 

For those who know the engine model itself a little better, the docstring for tfoper says the polytropic efficiencies are maximums for a given device (max. polytropic efficiency for the fan). This isn't the scaling used by NPSS, which uses design point scaling (max isn't necessarily at design point). Would this be easy to change/how would we do that? (probably @askprash or @aditeyashukla)

WyattGiroux:
As an update, I've been switching out the functions regardless of the answer to my above question and have been having trouble getting the tfoper to run. It seems that the derivatives going to 0 outside the map convex hull in certain places. To fix this, I'm going to try having the map generator default back to drela's model outside the convex hull of the map.

Any thoughts?

PR #94, opened 2024-10-21
Gradient based optimization tutorial

aditeyashukla:
Added code and markdown tutorial for Ipopt, finite diff optimizer


example/example_opt.jl
```diff
@@ -120,19 +120,26 @@ end
...
-lower      = [7.0 , 20000.0, 0.40, 10.0, 0.1, 0.1, 0.10,   0.10,   0.1,    0.1,     700.0,  6,      0]
-upper      = [12.0, 60000.0, 0.65, 40.0, 1.0, 1.0, 0.15,   0.15,   1.4,    1.0,     2000.0, 15,     10] 
+
+lower      = [9.0 , 0.53, 25.0, 10000.0, 0.65, 0.1,  0.125,    0.125,    0.9,   0.7, 1400.0, 10.0,  1.25, 2.98]
+upper      = [11.0, 0.60, 30.0, 10900.0, 0.85, 0.4,  0.15,   0.15,   1.3,   1.0, 1650.0, 15.0, 2.0, 3.02] 
+#initial    = [10.987, 0.598, 27.58, 10015.46, 0.85, 0.1587, 0.1449, 0.126, 1.0295, 0.993, 1649.359, 13.346, 1.8436, 3.0 ]
+initial    = [10.5, 0.57, 26.0, 10668.0, 0.7, 0.25, 0.1268, 0.1266, 1.238, 0.9, 1580.0, 12.0, 1.685, 3.0 ]
+
+
+# #             AR    Alt(ft)  Cl     Λ     λs  λt   hboxo   hboxs   rcls    rclt     Tt4CR   iepihc iepif
```
ngomezve, line 130: Why are you leaving the old params commented out? May be better to just delete them

example/optimization/Gradient_Optimizer_iterations.svg
ngomezve, line 1: This figure is the same as the one in the documentation. I'm not convinced that we need it twice.

docs/src/examples/gradient_based_optimization.md
```diff
@@ -0,0 +1,180 @@
...
+## Objective Function
+
+```julia
+function size_ac(x...)
+    # println("State: ",x)
+    ac = deepcopy(default_model)
+    # Set params
+    for (i,x_i) in enumerate(x)
+        field_path, index = params[i]
+        TASOPT.setNestedProp!(ac, field_path, x_i, index)
```
ngomezve, line 82: May be nice to add a comment explaining what this does. It may not be obvious that we need this convoluted syntax to access the par fields.

docs/src/examples/gradient_based_optimization.md
```diff
@@ -0,0 +1,180 @@
...
+include(__TASOPTindices__)
+```
+
+Initialize arrays used for plotting as well as the finite difference parameter
+
+```julia
+design_variables = []
+objective_array = []
+
+epsilon = 1e-5
```
ngomezve, line 37: Comment that this is the relative tolerance used in the FD method?

docs/src/examples/gradient_based_optimization.md
```diff
@@ -0,0 +1,180 @@
...
+            g[k] = grads
+        end
+    end
+end
+```
+
+## Running the optimization
+
+```julia
+function exampleOPT()
```
ngomezve, line 122: I think more comments on the IpOpt specific syntax would be nice, e.g., why you need to register functions

docs/src/examples/gradient_based_optimization.md
```diff
@@ -0,0 +1,180 @@
...
+
+    @NLobjective(model, Min, eval_f(x...))
+
+    optimize!(model)
+    println("""
+    termination_status = $(termination_status(model))
+    primal_status      = $(primal_status(model))
+    objective_value    = $(objective_value(model))
+    """)
+
```
ngomezve, line 145: I think it would be valuable to add constraints too. In my experience with IpOpt, setting up the constraints was harder than any other thing. I believe the syntax is of the type:
`@NLexpressions(model, begin
    q_1, size_ac(vars...)
    q_2, FuelVolCstr()
    q_3, SpanCstr()
    q_4, ClimbCstr()
    q_5, Tt3Cstr()
    q_6, TmetalCstr()
    q_7, MTOWCstr()
    q_8, FanDiamCstr()
    q_9, FanSizeCstr()
    q_10, FanWingCstr()
    end)`

ngomezve (changes requested):
This PR has a lot of potential! Thanks @aditeyashukla for your great work; something like this would have saved me weeks a few months ago. 

I think there are a few improvements that can make this even better. The main issue with the optimization example is that it is unconstrained (other than the box limits) and, therefore, not very useful. Adding a few constraints would turn it into a much more valuable example, as users would only need to change the design variables and constraints while keeping the same structure.

example/example_gradient_based_opt.jl
```diff
@@ -81,9 +124,15 @@ function exampleOPT()
...
 
+    function eval_g(x...)
+        return size_constraint(x...)
+    end
+
     register(model, :eval_f, size(initial)[1], eval_f, fdiff_all!; autodiff = false)
+    register(model, :eval_g, size(initial)[1], eval_g, fdiff_all_for_constraint!; autodiff = false)
 
     @NLobjective(model, Min, eval_f(x...))
+    @NLconstraint(model, eval_g(x...) >=0)
```
ngomezve, line 135: Shouldn't the constraint be that `eval_g`$\leq 0$? For example, with `(ac.parg[igWfuel]/ac.parg[igWfmax] - 1.0)` we want `ac.parg[igWfuel]<ac.parg[igWfmax]` so that  `(ac.parg[igWfuel]/ac.parg[igWfmax] - 1.0)`<0

example/example_gradient_based_opt.jl
```diff
@@ -68,6 +94,23 @@ function fdiff_all!(g, x...)
     end
 end
 
+function fdiff_all_for_constraint!(g, x...)
```
ngomezve, line 97: There must be a way we can avoid this. By doing this calculation, you are doubling the number of function evaluations. If you want to impose each constraint with a separate function, it would scale proportionally to the number of constraints. With `fdiff_all!` we are already perturbing the TASOPT aircraft (and all its outputs) by all the optimization variables. Talk to Dr. Ranjan about the best way to do this. Maybe you can save the value of the constraint derivatives in some temporary file or global variable?

ngomezve (changes requested):
@aditeyashukla Good improvements to the code! I think the constraint example will indeed help and the constraints you chose make sense. However, I believe that there are more efficient ways to impose the constraints than what is in this PR. The main issue is that this implementation doubles the number of function calls and I really don't think this is necessary.

aditeyashukla:
@ngomezve found a way to have the same number of function calls as no constraint - used a memorization function to "cache" the values for the same point in the design space

docs/src/examples/gradient_based_optimization.md
```diff
@@ -0,0 +1,201 @@
...
+# set_silent(model)
+set_optimizer_attribute(model, "tol", 1e-3)
+set_optimizer_attribute(model, "nlp_scaling_method", "gradient-based")
+set_optimizer_attribute(model, "obj_scaling_factor", 1.0)
+set_optimizer_attribute(model, "nlp_scaling_max_gradient", Float64(1))
+set_optimizer_attribute(model, "nlp_scaling_min_value", 1e-8)
+
+@variable(model, lower[i] <= x[i=1:length(initial)] <= upper[i], start=initial[i])
+
+@operator(model, f_size_ac, length(initial), memoized_size_ac[1],memoized_fd_ac[1])
```
ngomezve, line 185: Nice syntax improvement

ngomezve (approved):
@aditeyashukla Great job with storing the FD results so that wsize needs to be called the same amount of times for any number of constraints. I think this is a very solid example now. It would be nice to do some comparison between this implementation of IpOpt and the previous NLopt in terms of solution and optimization cost, but that can be done in the future.

PR #97, opened 2024-11-04
max payload tweak + follow-ups

argonaut22:
Edit in the `read_input` checks to include off-design missions in the max_payload exit_limit checks

context: discussed with @speth and thought about this framing vs. a max_payload in tons. decided against it since this is equivalent, though all of this needs documenting

### DIFF
--- a/src/IO/read_input.jl
+++ b/src/IO/read_input.jl
@@ -152,7 +152,7 @@ if any(maxpax .< pax)
     "\n Payloads listed [in pax] = "*string(pax))
 end
 if any(exitlimit .< pax)
-    error("One or more missions have higher payload than the prescribed exit limit!:"*
+    error("One or more missions have higher passenger counts than the prescribed exit limit!:"*
     "\n Exit limit [in pax] = "*string(exitlimit)*
     "\n Payloads listed [in pax] = "*string(pax))
 end
--- a/src/IO/read_input.jl
+++ b/src/IO/read_input.jl
@@ -145,10 +145,16 @@ maxpax = readmis("max_payload_in_pax_equivalent") #This represents the maximum a
 pax = readmis("pax")
 exitlimit = readmis("exit_limit")
 despax = pax[1] #Design number of passengers
-if despax > maxpax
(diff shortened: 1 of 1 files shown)


argonaut22:
@ngomezve, can you pls document the cabin sizing assumptions/process within the "Mission and Sizing" subsection? 
- Of special import is how that process relates to the mission parameters (e.g., answering "how do I use this if I'm designing something with variable length? variable radius? variable pax?"). 
- Also, some record of where the `exit_limit`s come from would be ideal
- Lastly, a clear answer would  be good on how this framework would handle a configuration with non-passenger "belly" cargo*. I think it should be a simple conversion of tonnage to the max payload while keeping the exit limit in the proper place, but please confirm that there aren't inconsistent assumptions in this interpretation

*for reference, 777-300ER has max Wpax = 154 000 lbs / 370 pax = 415 lbs/pax > 215 lbs/pax used for RJ and SA). meaning it can accommodate a fair bit of non-passenger cargo

ngomezve:
@argonaut22 Great suggestions! I'll get to it.

src/IO/read_input.jl
```diff
@@ -145,10 +145,16 @@ maxpax = readmis("max_payload_in_pax_equivalent") #This represents the maximum a
...
-elseif despax > exitlimit
-    error("Design mission has higher # of passengers than the exit limit!")
+
+if any(maxpax .< pax)
+    error("One or more missions have higher payload than prescribed maximum aircraft payload!:"*
+    "\n Max Payload [in pax] = "*string(maxpax)*
+    "\n Payloads listed [in pax] = "*string(pax))
+end
+if any(exitlimit .< pax)
+    error("One or more missions have higher payload than the prescribed exit limit!:"*
```
ngomezve, line 155: Maybe say "One or more missions have number of passengers than the prescribed exit limit"

ngomezve:
@argonaut22 I addressed your suggestions for better docs in PR #101. Feel free to make more suggestions!

PR #98, opened 2024-11-04
example flag style propagation + minor corrections

argonaut22:
Noting that the conversion of `pari` is near(ish), a consistent flag syntax would be ideal. 

@speth suggested a convention of `<verb>_<subject>` like `is_laminar` or `has_shaft` for Boolean parameters, PLUS a string-based option selection for parameters that have more options (e.g `icool` can be `opt_cooling` which will be under `engine`). I would also vote to add a prefix for flags, e.g., `flag_is_laminar`, for ease of searching and identification in the codebase.

Thoughts/objections on this @everyone (@aditeyashukla @ngomezve @askprash )? 
Making sure we all agree before trying to push beyond this sample `HXfun.jl`

### DIFF
--- a/src/engine/hxfun.jl
+++ b/src/engine/hxfun.jl
@@ -69,9 +69,9 @@ Structure containing the heat exchanger geometric and material properties.
 
 !!! details "💾 Data fields"
     **Inputs:**
-    - `fconc::Bool`: flag for concentric geometry (true: concentric ; false: rectangular)
-    - `frecirc::Bool`: flag for recirculation (true: recirculation ; false: no recirculation)
-    - `fshaf::Bool`: flag for whether HX contains shaf(true: shaft ; false: no shaft)
+    - `flag_is_concentric::Bool`: flag for concentric geometry (true: concentric ; false: rectangular)
+    - `flag_recirculates::Bool`: flag for recirculation (true: recirculation ; false: no recirculation)
+    - `flag_has_shaft::Bool`: flag for whether HX contains shaft(true: shaft ; false: no shaft)
     - `N_t::Float64`: number of tubes per row
     - `n_stages::Float64`: number of different coolant stages with different coolant flows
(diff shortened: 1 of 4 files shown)


argonaut22:
Should we choose to proceed, here's a probably-not-but-maybe-comprehensive list of replacements on deck. all are booleans unless marked by *:
- [ ] `HXfun.jl`
- [ ] `blsys` for `lam`, `wake`, etc. all boolean
- [ ] `cdsum`: `icdfun`
- [ ] `trefftz`: `Lspec`
- [ ] `balance`: `iiengloc`*, `iiHTsize`*, `iixwmove`*
- [ ] `tanksize`: `flag_size`, `flag_vacuum`
- [ ] `fuselage_tank`: `size_insulation`
- [ ] `tfan`: `icool`*
- [ ] `aircraft`: `sized` (also see #99)
- [ ] `size_cabin`: `flag_aisle`, `fdoubledecker`
- [ ] everything in `pari`

argonaut22:
As an example for how the string comparison would work, the comparison could be:
```
if engine.opt_cooling == "none"
    ...
elseif engine.opt_cooling == "fixed_BPRc"
    ...
elseif engine.opt_cooling == "fixed_Tmetal"
    ...
end
```

askprash:
This is good, we should do this! Personally, I'm not sure we need the `flag_` prefix - just makes it longer, and there's something nice about reading `if is_laminar` in the code but not a strong opposition. For the option string variables though having the `opt` prefix seems nice. 

Also for the string variables it would be good to `lowercase()` everything while testing to be consistent.

argonaut22:
Consensus is `flag_` not needed. Will incorporate the changes above in the next few days

askprash:
@argonaut22 is this one we can sort out for the release? It should also help get rid of some flags inside `pari`.

PR #101, opened 2024-11-06
Improved cabin sizing documentation

ngomezve:
This PR follows the request in PR #97 for better documentation on the cabin sizing models. A new page has been added to the `docs` summarizing the modeling approach for the single and double decker aircraft. In addition, old tests of the resizing functions that were removed when changing the fuselage data storage have been brought back.


src/IO/read_input.jl
```diff
@@ -142,8 +142,9 @@ if !ac_type_fixed
 end
 
 maxpax = readmis("max_payload_in_pax_equivalent") #This represents the maximum aircraft payload in equivalent number of pax
+                                                #Part of this payload may be carried in the aircraft's cargo hold
```
argonaut22, line 145: do you mean to clarify that maxpax includes tonnage beyond passenger-associated weight?

If so, consider rephrasing something like, "this may exceed the seatable capacity to account for belly cargo."

tangential question, is this extra mass assumed to distribute uniformly along the cabin? if so, seems worth noting in the docs (if not already done).

docs/src/structures/cabin_sizing.md
argonaut22, line 1: A good detailed pass. Two suggestions:
- a _very_ high level summary as the introductory paragraph would be ideal. Something that answers very quickly: "tasopt includes these assumptions by default, additional models can include y (e.g., doubledecking), you can calculate/optimize x if you specify blah". 
- On a more nitpicky stylistic note, the gen'l structure for each documentation page is: 1. high level summary, 2. conceptual/theoretical breakdown with figures and/or references (with links) to the relevant fxns 3. list of fxns as a list at the end. adhering to this i think would be helpful in keeping things consistent for readers

src/structures/size_cabin.jl
```diff
@@ -21,7 +22,7 @@ length.
 function place_cabin_seats(pax, cabin_width, seat_pitch = 30.0*in_to_m, 
     seat_width = 19.0*in_to_m, aisle_halfwidth = 10.0*in_to_m, fuse_offset = 6.0*in_to_m)
 
-    cabin_offset = 10 * ft_to_m #Distance to the front and back of seats
+    cabin_offset = 10 * ft_to_m #Distance to the front of seats
     #TODO the hardcoded 10 ft is not elegant
```
argonaut22, line 26: agreed. off-the-cuff thought: perhaps it could be a default value for a field within the Cabin struct (should that be fully implemented)

argonaut22 (commented):
overall, 👍. a couple of comments within which should help with consistency

src/IO/read_input.jl
ngomezve: I didn't make changes to the way that the maximum payload is handled by 'fuseW()'. My understanding is that it is indeed uniformly distributed.

docs/src/structures/cabin_sizing.md
ngomezve: Thanks for the suggestion. I just made these changes.

argonaut22:
Thanks for making the changes. Made a few more and i think we're good to go.

argonaut22 (approved):

PR #102, opened 2024-11-06
Update README.md to fix CI badge

askprash:
CI badge keeps showing failing cause the test_depr one has failures - trying to be specific which test we want to show on the badge


askprash:
I think we might want to split out the strict deprecation test to a different yml file.

PR #104, opened 2024-11-08
Introducing models of ducted fan propulsion

ngomezve:
This PR introduces a new propulsion architecture to TASOPT: ducted fans. These fans are assumed to be powered by a system that can provide mechanical power (e.g., an electric motor). Models for motors and cables are currently in development by @askprash, but have not been pulled in yet.

The ducted fan models are based on the existing turbofan models, but the core has been removed. Two functions are introduced: `ductedfansize!()` sizes the ducted fan for a desired mission point and `ductedfanoper!()` calculates the off-design performance for specified power or thrust. 

The function `ductedfanoper!()` solves a small non-linear problem by using the NLsolve.jl suite. This is not as efficient as the direct Jacobian computation in `tfoper!()` but is easier to modify.


src/engine/ducted_fan/ductedfancalc.jl
```diff
@@ -0,0 +1,287 @@
...
+
+    pare[ieTt18] = Tt18
+    pare[ieht18] = ht18
+    pare[iept18] = pt18
+    pare[iecpt18] = cpt18
+    pare[ieRt18] = Rt18
+
+    pare[ieTt2] = Tt2
+    pare[ieht2] = ht2
+    pare[iept2] = pt2
```
aditeyashukla, line 238: not urgent to fix but a lot of this file is just packing/unpacking of variables (similar to tfcalc) - I think if we get rid of pare (especially for newer updates) we can shortern these files

src/engine/fuel_cell/FC_objects.jl
aditeyashukla, line 1: can we rename the file to something else (probably need a nomenclature for such files for the entire repo)

src/engine/thrust_from_ROC.jl
aditeyashukla, line 1: do you think its worth adding a test for this file? or not important

aditeyashukla (approved):
Very cool update - I think this has great future use. Made some comments mostly about practice and nomencalture moving forward as we add new technology - but since we should probably discuss it in meeting first, I'd say this is ready to be pulled in

src/engine/ducted_fan/ductedfancalc.jl
ngomezve: Agreed

src/engine/fuel_cell/FC_objects.jl
ngomezve: Good idea for the future

src/engine/thrust_from_ROC.jl
ngomezve: This was a great suggestion! I just caught a bug brought over from the recent change in the unpacking function that the PR did not include. I added a simple test for the function.

aditeyashukla (approved):

PR #105, opened 2024-11-14
Heat exchanging: radiator, icing & better optimization

ngomezve:
This PR introduces models of radiators to reject residual heat, which would be required by electric systems with batteries or fuel cells. It reuses the existing heat exchanger models and adds new functions for the design and analysis of radiators.

In addition, the following changes are made:
- The heat exchanger optimization now uses the COBYLA algorithm instead of Nelder-Mead with penalty functions.
- The minimum heat exchanger temperature at every design point is stored to check for icing.


askprash:
@ngomezve looks like the this is failing in the Julia 10.6 version - I must have missed it before pulling in. Could you please fix?

[See here.](https://github.com/MIT-LAE/TASOPT.jl/actions/runs/12299353138/job/34324946261)

PR #107, opened 2024-11-20
Adding landing gear mass models

ngomezve:
This PR introduces new landing gear mass models based on historical correlations. In the future, physics-based models could be included as well. The following changes are made:
- Two landing gear models are supported: the previous constant mass fractions and a model based on historical data, which incorporates dependencies on landing gear length and design.
- The landing gear parameters are stored in a new object in `ac`. The related `parg` parameters are eliminated.
- The balance() and mission!() functions now take in the `ac` object directly instead of the `par` arrays; these are then unpacked inside the functions. This makes the syntax more robust to changes in data storage.
- Documentation and tests for the landing gear models.


ngomezve:
Superseded by PR #128

Issue #106, opened 2024-11-17
Payload-Range capabilities need updating 


argonaut22:
Updated after rediscovering the outputs.jl `PayloadRange()`:

steps to bring the function to best practices:
- [x] create new tests to ensure fixes for future PRs
- [x] Update plotting to use Plots.jl (done in branch: argonaut22/plot_redux)
- [ ] can we do better* than try-catching any error that gets thrown? the effect with this handling is that any breaking gets hidden and a 0-range diagram is produced with little indication of what went wrong.

With PayloadRange(), we can probably delete `example_PRD.jl` and `PRD_input.toml`. If we somehow wanted to keep it, the following would need to be done, since it's currently broken:
- [ ] update `PRD_input.toml` to match the new max_pax nomenclature, and aircraft_type field
- [x] why does the PRD toml example specify 5 missions?

*we should do better 😬

argonaut22:
Update! found the version in `outputs.jl`. adjusted tasks accordingly

ngomezve:
I'll fix this in my PR for off-design missions. The `PayloadRange()` function is quite outdated and requires some specific aircraft objects (e.g., it cannot have just one mission).

argonaut22:
Fun new issue: running `PayloadRange` on an aircraft with only one mission does not work (tries to set ac.parm[imRange,2] when ac.parm is only a Vector). Can replicate with the `example_cryo.jl` model.

![Image](https://github.com/user-attachments/assets/70aec1e2-9246-418a-b83d-733e1e547574)

argonaut22:
After discussing with @askprash , best path forward seems to be to add some code to the start of `PayloadRange()` to:
- copy/`unpack_ac` the requested `aircraft` into a new one with _two_ missions: the design one, and a dummy one
- the dummy mission will be edited by the payload-range study/fxn and exercised with `fly_off_design()`

This approach addresses the above bug, ensures that the original aircraft does not get accidentally edited, and prevents having to add/remove a mission from an extant model (a convenience that we may want to provide, but def not a priority).

PR #73, opened 2024-08-05
Refactoring the wing and tail sizing 

aditeyashukla:

PR #109, opened 2024-12-04
Reviving woper() for off-design performance

ngomezve:
This PR modifies the old `woper()` function for the off-design aircraft performance; the function is renamed to `fly_off_design()`. This function calculates the fuel weight needed to perform missions different to those in the design mission. The new fuel, PFEI and takeoff weights get stored in `parm`.

 `woper()` had not been maintained for a while and it was no longer operational. It also had some hardcoded parameters that have been removed. The functions to calculate fuel venting if the fuel is cryogenic have been introduced as well.

This PR also addresses issue #106.  The payload-range diagram function and examples have also been improved: the function now also plots PFEI for maximum payload, and the grid can be non-uniform with greater density near the design range and maximum payload. This function now also works with aircraft that only have one mission.


ngomezve:
I just updated this PR to remove the conflicts after merging #73

askprash (approved):
LGTM!

PR #113, opened 2024-12-16
Generalized propulsion functions

ngomezve:
This PR introduces new engine functions: `enginecalc!` and `engineweight!` that replace the engine function calls. These functions act as handles for any propulsion function that can use the same syntax, e.g., do design or off-design cases.

The `enginecalc!` and `engineweight!`  functions are stored and then extracted from a new `Engine` object produced when the input file is read. The user is free to then replace those functions with others if a more custom propulsion option is desired.

The old turbofan functions are now stored in a dedicated `turbofan` folder inside `src/engine`. New functions called `tfwrap!` and `tfweightwrap!` have been created as wrappers for the old functions and related pieces of code, such that these can be called by the generalized functions.


src/misc/engine.jl
```diff
@@ -0,0 +1,63 @@
...
+"""
+@kwdef mutable struct Engine
+    """Engine performance model identifier"""
+    model_name::String = ""
+    """Engine function to be used by TASOPT"""
+    model::Function = (x->x)
+    """Weight model identifier"""
+    weight_model_name::String = ""
+    """Weight model to be used by TASOPT"""
+    weight_model::Function = (x->x)
```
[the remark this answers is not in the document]
ngomezve: Thanks for the comments! Will get started on the updates.

ngomezve:
I made the changes suggested by @askprash. The code is still somewhat slower and with more allocations than `main`; not totally sure why. In `main`, the default aircraft median sizing time is 62.773 ms with 663255 allocations. In `dev_generalprop`, the median time is 64.327 ms with 706241 allocations.

src/sizing/wsize.jl
argonaut22, line 1: good streamlining. only note, `case` might be a good variable to start with the string option style (i.e., `opt_case`)

src/engine/engine.jl
```diff
@@ -8,11 +8,13 @@ using NLopt
 using Roots
 using LinearAlgebra
 
-export tfcalc!, mcool, Tmcalc, gas_tset, gaschem
-export tfweight, ddct, ddat, gct, gat, tfsize!, Ncmap, ecmap, Ncmap1, ecmap1, etmap, Pimap, tfoper!
+export Engine, store_engine_model!, extract_engine_model
+
+export tfwrap!, tfcalc!, mcool, Tmcalc, gas_tset, gaschem
+export tfweightwrap!, tfweight, ddct, ddat, gct, gat, tfsize!, Ncmap, ecmap, Ncmap1, ecmap1, etmap, Pimap, tfoper!
```
argonaut22, line 14: do we actually need all of these vars exported to the main namespace? I see that they were there before, but may as well prune if they're not needed

src/misc/engine.jl
```diff
@@ -0,0 +1,47 @@
...
+
+Engine and models
+
+$TYPEDFIELDS
+"""
+mutable struct Engine{M<:AbstractModel}
+    model::M
+
+    """Heat exchanger parameters and data"""
+    heat_exchangers::Vector{HX_struct}
```
argonaut22, line 14: can we do this without the heat exchangers in every model? i think the superclass should be as general as possible, and it's not clear to me/users why `HX_struct`s have a special position.

src/engine/turbofan/tfwrap.jl
```diff
@@ -0,0 +1,89 @@
...
+turbofan model. This function is basically a wrapper on tfcalc!, going from the
+basic engine inputs to those required by the function and storing the outputs.
+
+!!! details "🔃 Inputs and Outputs"
+    **Input:**
+    - `ac::aircraft`: aircraft object
+    - `case::String`: case identifier, e.g. "sizing" or "off_design"
+    - `imission::Int64`: mission index
+    - `ip::Int64`: mission point index
+    - `initeng::Int64`: flag to initialize engine
```
argonaut22, line 14: getting ahead of the flag style PR, can we rename `initeng::Int64` to `initializes_engine::Bool`

src/engine/turbofan/tfwrap.jl
```diff
@@ -0,0 +1,89 @@
...
+        0  initialize variables for iteration in engine
+        1  use current variables as initial guesses in engine
+    - `iterw::Int64`: sizing loop iteration
+
+    **Output:**
+    No direct outputs. The `ac` object gets modified with the engine parameters.
+"""
+function tfwrap!(ac, case::String, imission::Int64, ip::Int64, initeng::Int64, iterw::Int64 = 0)
+    #Unpack data storage arrays
+    pari = ac.pari
```
argonaut22, line 24: `pari, parg, _, para, pare, _, _, wing, _, _, _= unpack_ac(ac)` ?

src/engine/turbofan/tfwrap.jl
```diff
@@ -0,0 +1,89 @@
...
+        icool = 1
+        if (iterw == 1 || initeng == 0)
+            # initialize engine state
+            inite1 = 0
+        else
+            # start with current engine state
+            inite1 = 1
+        end
+
+        ichoke5, ichoke7 = tfcalc!(pari,parg,view(para, :, ip), view(pare, :, ip), wing, ip, icall, icool, inite1)
```
argonaut22, line 41: again, flag style preemption: `inite1` -> `initializes_engine_firstiter` ?

src/engine/turbofan/tfwrap.jl
```diff
@@ -0,0 +1,89 @@
...
+
+        pare[iepifD, :] .= pare[iepifD, ip]
+        pare[iepilcD, :] .= pare[iepilcD, ip]
+        pare[iepihcD, :] .= pare[iepihcD, ip]
+        pare[iepihtD, :] .= pare[iepihtD, ip]
+        pare[iepiltD, :] .= pare[iepiltD, ip]
+        
+    elseif case == "off_design"
+        icall = (ip in range(ipstatic, ipclimbn)) ? 1 : 2 #One if in range, 2 if not
+        icool = 1
```
argonaut22, line 72: again again, flag style preemption:. this one is a bit more openended since we haven't yet propagated string-based `opt`s
- `icall::Int64` -> `opt_call::String` can be `"sizing", "oper_Tt4", "oper_Fe"`
- `icool` -> `opt_cooling::String` can be `"none", "fixed_flowratio", "fixed_Tmetal"`

argonaut22 (changes requested):
Great work overall. Generalization has been an open question for a bit, and big fan of unpack_ac. 

Left a few notes centering around:
1. flag style propagation (secondary, but we should get rolling on this)
2. including HX in the most general functions seems unintuitive
3. misc tweaks

src/engine/engine.jl
ngomezve: Good catch; some of those functions no longer exist

src/IO/read_input.jl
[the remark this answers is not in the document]
ngomezve: That's a good addition and something that we should definitely do in the future. For this PR, I wanted to focus on the changes to the code while minimizing changes to the inputs, but definitely something to consider.

src/misc/engine.jl
ngomezve: The thought is that there will be more engine elements and the HX one is the only one implemented right now. In the future we could have turbomachinery as well.

ngomezve: I added a todo comment to clarify it

src/engine/turbofan/tfweightwrap.jl
```diff
@@ -0,0 +1,42 @@
...
+      
+!!! details "🔃 Inputs and Outputs"
+    **Input:**
+    - `ac::aircraft`: aircraft object
+    - `HXs`: vector with heat exchanger performance data
+
+    **Output:**
+    No direct outputs. The `ac` object gets modified with the engine weights.
+"""
+function tfweightwrap!(ac, HXs)
```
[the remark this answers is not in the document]
ngomezve: Agreed and done

src/engine/turbofan/tfwrap.jl
argonaut22: oh interesting. i had not considered build order to this point, but makes sense

I got it to work by sticking the fxns in `misc/unpack_ac.jl` and building before `include(__TASOPTindices__)`. thoughts on this strategy? @askprash ? 

i can commit and push if we like it

src/engine/turbofan/tfwrap.jl
ngomezve: Done

src/engine/turbofan/tfwrap.jl
ngomezve: I made the change although I'm not convinced that the syntax is much better

src/engine/turbofan/tfwrap.jl
ngomezve: This is potentially a good improvement but it would require changes to `tfcalc!`. My goal with this PR is lay out a data structure that could work with any function so I'd rather avoid this for simplicity now.

argonaut22: if it's not in this PR, it'll be in #98 next so the changes have to happen anyway. 

Given your familiarity, do you foresee any issues or is it only going through ctrl-f?

src/engine/turbofan/tfwrap.jl
ngomezve: Fixed the issue by splitting `misc/aircraft.jl` into two files so that the helper functions can be loaded separately.

src/engine/turbofan/tfwrap.jl
ngomezve: Probably no major issues. We'll probably have to adapt the syntax in `wsize()` to accommodate the change in flags, but I don't anticipate this breaking anything.

argonaut22 (approved):

argonaut22 (approved):

argonaut22 (approved):

Issue #74, opened 2024-08-06
Rewrite test/unit_test_io.jl with new fuselage/wing/tail


aditeyashukla:
@argonaut22 IO unit test doesn't work because of new fuselage (and soon wing/tail) 
we can fix this when wing/tail is pushed as you would have to fix it then too (If you want to get a head start you can pull the branch `wing_remodel` on my branch

askprash:
Left some suggestions and caveats here:
https://github.com/MIT-LAE/TASOPT.jl/blob/d2d9b340290b0acfccc5e8818f4ef2b04b11fd39/src/IO/quicksave_load.jl#L22-L26

and here:

https://github.com/MIT-LAE/TASOPT.jl/blob/d2d9b340290b0acfccc5e8818f4ef2b04b11fd39/src/IO/quicksave_load.jl#L108-L123

PR #114, opened 2024-12-16
Plots.jl Refactor (goodbye PythonPlot)

### LINKED ISSUE #103: Updating Plotting and Data Functionalities
Python-dependent plotting increases installation footprint and likelihood of package-management woes. `Plots.jl` is a lightweight way to do the same thing and largely follows the `PyPlot` conventions. I'm thinking we should replace this before it gets more complicated.

Moreover, @speth votes for limiting plotting within TASOPT itself for diagnostics, relying on external plotters for more involved stuff. 

I'll address this and #74 in the same PR to support this paradigm^

argonaut22:
This PR:
- surgically excises the metastasized plt usage (so we're not installing a GB of python just to plot)
- replaces it in all places with plt-like plotter, Plots.jl (see `outputs.jl` for extensive sample usage)
- Splits `outputs.jl` into `text_outputs.jl` and `plot_outputs.jl`
- ensures all of this is tested such that changes to data structure will show a break
- closes #103. noting that the show_plot/verbosity parameter is reserved for another PR

Note:
- some TODO's are marked where the originally-intended functionality appears absent. Of note: `label_bars()` which is now replaced with simple legends in the plot
- data IO tests are re-instated, broken tests are appropriately marked. some are /still/ commented out since they're dependent on the data struct handling task, but they're marked as TODO
- there may be a better plotting backend to select that would allow direct copying from VScode/etc. Unclear what that is rn

Sample outputs:
image
image


example/example_PRD.jl
aditeyashukla, line 1: guessing this will have to change with Nico's PR about the payload range/woper updates

aditeyashukla (commented):
Looks great! 
Will need updates once Wing/tail PR and Woper PR are merged
Happy to help for updates regarding the former

src/IO/outputs.jl
```diff
@@ -463,161 +463,175 @@ function stickfig(ac::aircraft; ax = nothing, label_fs = 16,
...
+
+    if annotate_length
+        yloc = -1.1*maximum(yw)
+        plot!([0.0, xf[end]], [yloc, yloc], lw=1.5, arrow=:h, color=:black, label="")
+        annotate!(xf[end] *0.25, yloc+1.5, text("\$\\ell\$ = $(round(xf[end], digits=1)) m",
+            halign=:center, valign=:center, fontsize=14))
+    end
+
+    # Span annotations
+    find_aerodrome_code(bmax) = (["C", "III"], bmax)  # Replace with actual function
```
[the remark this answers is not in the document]
argonaut22: good eye. the function is still here (in this version, line 624) but looks like a placeholder overstayed its welcome. will fix

argonaut22:
Rebased onto b6129e2 (from #73)
propagating wing changes through to `save_model.jl` and other overlooked files

argonaut22:
Replaced some straggler references to wing parg[] vars. 

Marked newly-broken IO tests which will be fixed as another PR. Aiming to do this before additional struct introductions so that the variable changes can be done at once and systematically.

Review and pull before other stuff changes plssss

argonaut22:
Also of note:
- found some inconsistencies between the code and the var name spreadsheet. made changes to the spreadsheet. be on the lookout
- #74 is still pending, though the triaging for it is done here

src/IO/output_plots.jl
ngomezve, line 1: In the future, consider using `git mv` to rename files when you are keeping most of the content. This way, the diffs are properly tracked.

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1028 @@
...
+    scatter!(plot_obj, [parg[igxNP]], [0.0], color=:blue, marker=:circle, z_order=15, label="NP")
+    annotate!(plot_obj, (parg[igxNP]+2, 0.5, text("NP", label_fs - 2.0, :center, :center, color=:blue)))
+
+    # Annotate CG range
+    plot!(plot_obj, [parg[igxCGfwd], parg[igxCGaft]], [0.0, 0.0], color=:black, lw=2.0, label="CG movement")  # Line between points
+    scatter!(plot_obj, [parg[igxCGfwd], parg[igxCGaft]], [0.0, 0.0], marker=:vline, color=:black)                # End points
+    annotate!(plot_obj, parg[igxCGfwd]-2, 0, text("CG", label_fs - 2.0, :center, :center))
+
+    # Show seats (single-deck case)
+    if pari[iidoubledeck] == 0
```
ngomezve, line 415: Not all seats are currently being shown. Is this by design? It would be good to see if aircraft are single or twin aisle

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1028 @@
...
+
+    # Tank plotting
+    if pari[iifwing] == 0
+        for m in 1:nftanks
+            # Tank outline
+            plot!(plot_obj, xt[:, m], yt[:, m], color=:black, lw=1.5, z_order=10)
+            plot!(plot_obj, xt[:, m], -yt[:, m], color=:black, lw=1.5, z_order=10)
+            
+            # Filled area between
+            plot!(plot_obj, xt[:, m], yt[:, m], fillrange=-yt[:, m], color=:red, alpha=0.1, z_order=6, linewidth=1.0)
```
ngomezve, line 344: The tank is currently not being shown. Please run `example/example_cryo.jl` to see an example hydrogen aircraft and how it looks.

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1028 @@
...
+    - `ac_og::aircraft`: Aircraft structure for payload range diagram.
+    - `Rpts::Int64`: Density of ranges to be plot (Optional).
+    - `Ppts::Int64`: Density of payloads to be plot (Optional).
+    - `filename::String`: filename string for the plot to be stored (Optional).
+    - `OEW::Boolean`: Whether to have OEW+Payload on the y-axis or just Payload (Optional).
+    - `itermax::Int64`: Max Iterations for woper loop (Optional).
+    - `initeng::Boolean`: Use design case as initial guess for engine state if true (Optional)
+"""
+
+function PayloadRange(ac_og; Rpts = 20, Ppts = 20, filename = "PayloadRangeDiagram.png", OEW = false)
```
ngomezve, line 955: This function will be extensively modified in PR #109. There will be conflicts...

src/IO/output_texts.jl
ngomezve, line 1: Good split! Outputs was unnecessarily long.

ngomezve (changes requested):
Good PR. PythonPlots is very difficult to install properly. There are a few issues with functionality that we used to have that seems to be broken; for example, LH2 tanks not showing or only some seats being displayed.

src/IO/output_plots.jl
argonaut22: The primary value-add here is the plotting. everything else is broadly as it was. 

Thus, i don't think it'll be too bad to merge so long as the plotted variables remain the same.

src/IO/output_plots.jl
argonaut22: looking into it

src/IO/output_plots.jl
argonaut22: would you mind dropping a screenshot here? Troubleshooting `PythonPlot` is anathema to me after all this

src/IO/output_plots.jl
argonaut22: et voila

image

src/IO/output_plots.jl
askprash: Double check the z_orders

argonaut22: indexing error turns out. `xt[m, :]`, not `[:,m]`

example/example_PRD.jl
argonaut22: just a matter of switching the function name, huzzah

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1053 @@
...
+            halign=:left, valign=:top, color=:black),
+            z_order=:front)
+
+        
+        plot!(plot_obj, right_margin=200px) #"left-aligns" the plot within the figure
+    end
+
+    if annotate_length
+        yloc = -1.1*maximum(yw)
+        plot!(plot_obj, [0.0, xf[end]], [yloc, yloc], lw=1.5, arrow=:h, color=:black, label="")
```
ngomezve, line 459:
```suggestion
        plot!(plot_obj, [0.0, xf[end]], [yloc, yloc], lw=1.5, arrow=:both, color=:black, label="")
```

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1053 @@
...
+
+    # Tank plotting
+    if pari[iifwing] == 0
+        for m in 1:nftanks
+            # Tank outline
+            plot!(plot_obj, xt[m,:], yt[m,:], color=:black, lw=1.5, z_order=10)
+            plot!(plot_obj, xt[m,:], -yt[m,:], color=:black, lw=1.5, z_order=10)
+            
+            # Filled area between
+            plot!(plot_obj, xt[m,:], yt[m,:], fillrange=-yt[m,:], color=:red, alpha=0.1, z_order=4, linewidth=1.0)
```
ngomezve, line 357:
```suggestion
            plot!(xt[m, :], yt[m, :], fill_between=(-yt[m, :], yt[m, :]), color=:red, alpha=0.1)
```

ngomezve: The original code does not work but the suggestion should.

src/IO/outputs.jl
```diff
@@ -1,1219 +0,0 @@
...
-    else #greater density around design range
-        RangeMap = unique([LinRange(0.01,1, Int64(ceil(Rpts/2)+1)).^0.5; 1 .+ LinRange(0,1, Int64(floor(Rpts/2))).^2])
-        PayMap = LinRange(1, 0, Ppts).^0.75 #Use a map with greater density in the high payloads
-    end
-    RangeArray = ac.parm[imRange,1] * RangeMap
-    Payloads = PayMap * maxPay
-
-    RangesToPlot = []
-    PayloadToPlot = []
-    PFEIsToPlot = []
```
ngomezve, line 1153: Why did you remove the PFEI plot? I think it is a nice addition to the payload-range plot...

ngomezve (commented):
Good job improving the PR. I think it is almost there. I made a couple of small suggestions on how to improve the stickfig so that the length arrow and the tank color fill show properly. 

On the payload-range diagram, this PR removes some functionality I added on the previous one, namely the PFEI plot is no longer shown. It would be nice if this were added again.

src/IO/outputs.jl
ngomezve: Is the behavior on the drag and mass details expected? I believe the subplots need to be adjusted.

Example

src/IO/output_plots.jl
argonaut22: good catch. dunno what happened there

argonaut22:
Thanks Nico. Fixed:
image
image

The empty weight breakdown raises an eyebrow

test/unit_test_io.jl
```diff
@@ -17,25 +17,32 @@
...
+    #until then, these need to be disabled
+
+    
     #test that quicksave/load roundtrip default aircraft sizes identically to default load
     filepath_quick = joinpath(TASOPT.__TASOPTroot__, "../test/iotest_quick.toml")
-    quicksave_aircraft(load_default_model(), filepath = filepath_quick)
+    quicksave_aircraft(load_default_model(), filepath_quick)
     ac_quick = quickload_aircraft(filepath_quick)
-    size_aircraft!(ac_quick, Ldebug=false, printiter=false, saveOD=false)
+    size_aircraft!(ac_quick, Ldebug=true, printiter=true, saveOD=false)
```
ngomezve, line 45:
```suggestion
    size_aircraft!(ac_quick, Ldebug=false, printiter=false, saveOD=false)
```

src/IO/output_plots.jl
```diff
@@ -0,0 +1,1079 @@
...
+    - `ac_og::aircraft`: Aircraft structure for payload range diagram.
+    - `Rpts::Int64`: Density of ranges to be plot (Optional).
+    - `Ppts::Int64`: Density of payloads to be plot (Optional).
+    - `filename::String`: filename string for the plot to be stored (Optional).
+    - `OEW::Boolean`: Whether to have OEW+Payload on the y-axis or just Payload (Optional).
+    - `itermax::Int64`: Max Iterations for fly_off_design! loop (Optional).
+    - `initeng::Boolean`: Use design case as initial guess for engine state if true (Optional)
+"""
+
+function PayloadRange(ac_og::TASOPT.aircraft;
```
ngomezve, line 978: Add verbosity switch to prevent long outputs in tests

src/IO/quicksave_load.jl
ngomezve, line 1: Nice addition!

ngomezve (commented):
Very good improvements once again. Only issue is that the test file is too verbose because of unnecessary outputs. Probably fixable in one commit.

ngomezve (approved):
Approving PR after last changes

ngomezve (approved):
Approving PR. Thanks @argonaut22 for the improvements.

PR #119, opened 2025-01-23
Fuselage fuel tank cleanup

ngomezve:
This PR simplifies the structure of the code used to size and calculate the weight of fuel tanks in the fuselage. The previous code retained functionality that is no longer needed (e.g., calculating constant boiloffs when now there is a dedicated boiloff model) and was more cumbersome than it needs to be.

The PR moves some pieces of code from `wsize()` to `tank_size()` and `update_fuse!()` and also changes the inputs of these functions so that they only need the `ac` object. 

The changes introduced in the PR are to the code structure only and should not change the underlying models.


src/structures/update_fuse.jl
```diff
@@ -1,55 +1,82 @@
...
-    - `parg::Vector{Float64}`: vector with aircraft geometric and mass parameters
+    - `ac::aircraft`: object with aircraft parameters
+    - `imission::Int64`: mission index
 
     **Outputs:**
-    No direct outputs; parameters in `parg` are modified.
+    No direct outputs; parameters in `ac` are modified.
 """
-function update_fuse!(fuselage, wing, htail, vtail, pari, parg)
+function update_fuse!(ac, imission::Int64)
```
aditeyashukla, line 15: small (and honestly ignorable) comment but you could add the default value for `imission` as 1 in the function definition so you can call `update_fuse!(ac)` in wsize

src/IO/output_plots.jl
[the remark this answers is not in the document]
ngomezve: Kind of. This was accidentally removed in PR #114 but I think it is valuable functionality. It is particularly useful when testing the H2 sample aircraft for off-design missions.

ngomezve:
This commit fixes a bug in which the tank sea level temperature was not being updated in off-design. It also fixes a minor bug if the aircraft is a double-decker.

src/structures/update_fuse.jl
ngomezve: Just did this. Also did the same thing for `tanksize!`.

askprash (approved):

PR #121, opened 2025-02-01
tweak CI.yml tests '1' for latest stable release instead of fixed 1.11

argonaut22:


argonaut22:
rather than testing on 1.11 every time, setting to 1 uses the latest **stable** version every time

(started as an accidental PR from tests, but this teeny change seems worth the click.)

askprash (approved):

PR #122, opened 2025-02-04
struct support for output_csv

argonaut22:
`output_csv` can now print structs and fields. Note that by default these are A LOT of fields (e.g., 246 from `material`s). Fields can be excluded via keyword argument. ex:

> `output_csv(ac, struct_excludes = ["material", "fuse_tank"]`

With this and the IO changes in #114 , IO features and tests are now back online after changes to par arrays. Adjustments will need to happen for when mission- and flight-point-dependent par arrays are killed, as ID'd by the tests. More trivially, a couple of tests will need tweaks when new structs are introduced/par arrays are killed.


src/IO/output_csv.jl
```diff
@@ -145,11 +162,41 @@ function output_csv(ac::TASOPT.aircraft=TASOPT.load_default_model(),
...
+    field_values = [getfield(s, f) for f in fieldnames(T)]
+
+    flat_names = []
+    flat_values = []
+    
+    #iterate over field-value pairs
+    for (name, value) in zip(field_names, field_values)
+        full_name = prefix * name #add prefix to name if this struct is a field of another struct
+        #if value is a struct, call recursively
+        if isstructtype(typeof(value)) && !(value isa AbstractArray)
```
ngomezve, line 186: Nice recursive structure

ngomezve (approved):
Good PR. The exact use case of the output_csv function is still not totally clear to me but in any case, it is working.

argonaut22:
thanks! also lol, sweeps nico. programmatic sweeps

PR #127, opened 2025-02-11
LTO,PTF functionality and tests added

aditeyashukla:
Updated functions for 
- odperf
- LTO
- printBADA

Added P3T3 function for both narrow body (CFM56) and wide body (GE90) using curve fitting

Added tests for all 3 functions

Also, added `aircraft_type` in aircraft struct


src/mission/LTO.jl
aditeyashukla, line 1: rename to generate_BADA()

src/IO/read_input.jl
```diff
@@ -1060,7 +1060,7 @@ dHEx = dprop["HeatExchangers"]
     pare[ieTurbCepsilon, :, :] .= read_input("turbine_cooler_effectiveness", HEx, dHEx)
     pare[ieTurbCMp, :, :] .= read_input("turbine_cooler_inlet_mach", HEx, dHEx)
 
-return TASOPT.aircraft(name, description,
+return TASOPT.aircraft(name, description, aircraft_type,
```
aditeyashukla, line 1063: rename to something cooler

aditeyashukla: "clear what its for"

argonaut22:
To clarify my drowsy thoughts from earlier:
- I think `LTO.jl` and `printBADA.jl` belong in `IO/` as output_offdesign.jl: `output_LTO_perf` and `output_BADA`. Though there are calcs being done, they don't seem to affect the aircraft data and, thus, it's output
- `odperf` does seem to affect the ac via `tfcalc!` and `cdsum!`. thus, makes sense in `mission/`. Does running `odperf` change the aircraft for subsequent evaluations of other missions? I'm not that familiar with this functionality
- I still don't like `aircraft_type` as a top-level option since it's unused except for this one function. Would prefer a required parameter like `odperf!(ac, aircraft_type("narrow body"))`. If we _must_ have it, then please just use Strings. we evolved past the need for `Int` flags

src/IO/read_input.jl
```diff
@@ -41,6 +41,19 @@ function get_template_input_file(designrange)
...
+
+function ac_sym(aircraft_type::String)
+    if lowercase(aircraft_type) == "narrow body aircraft"
+        return :narrow
+    elseif lowercase(aircraft_type) == "wide body aircraft"
+        return :wide
+    elseif lowercase(aircraft_type) == "regional aircraft"
+        return :regional
+    else
+        return :narrow
```
argonaut22, line 53: how about `else @error(...)`?

src/IO/output_LTO_perf.jl
```diff
@@ -0,0 +1,141 @@
...
+        e =  -6.69412110e+00
+    end
+
+    H = -19.0*(sp_humidity - 0.00634)
+
+    EI_NOx = exp(H)*P3_kPa^0.4 * (a*T3_K^4 + b*T3_K^3 + c*T3_K^2 + d*T3_K + e)
+    return EI_NOx
+end  # function EINOx4
+
+function EINOx(ac::aircraft, ip::Int; sp_humidity = 0.00634, method="cubic")
```
argonaut22, line 132: This is the only place that uses `aircraft_type`. Would still vote against having it as a field and vote for having it as an input whenever `odperf!()` is called (which then passes to `EINOx` and co).

```suggestion
function EINOx(ac::aircraft, ip::Int, aircraft_type::Sym; sp_humidity = 0.00634, method="cubic")
```

argonaut22: the motivation is that "narrow-body"/"wide-body" is too big of a user-facing lever that does nothing anywhere else. it's unintuitive

src/IO/output_LTO_perf.jl
```diff
@@ -0,0 +1,141 @@
...
+"""
+    EINOx3(P3_kPa,T3_K, sp_humidity = 0.00634)
+
+
+Returns the EI(NOₓ) for a CFM56 level engine.
+Uses a third order polynomial in T₃ 
+Assumes a default specific humidity of 0.00634 kg water/kg dry air per 
+ICAO Annex 16 Vol. II (part 2.1.4.1)
+"""
+function EINOx3(P3_kPa, T3_K, sp_humidity = 0.00634, ac_type = "1")
```
argonaut22, line 82: ac_type default is outdated

Also, should this even have a default? the likelihood of error is greater than the convenience

src/IO/output_LTO_perf.jl
```diff
@@ -0,0 +1,141 @@
...
+end
+"""
+    EINOx4(P3_kPa, T3_K, sp_humidity = 0.00634)
+
+Returns the EI(NOₓ) for a CFM56 level engine.
+Uses a fourth order polynomial which behaves a little better at low T₃ than the cubic
+Assumes a default specific humidity of 0.00634 kg water/kg dry air per 
+ICAO Annex 16 Vol. II (part 2.1.4.1)
+"""
+function EINOx4(P3_kPa, T3_K, sp_humidity = 0.00634, ac_type = "1")
```
argonaut22, line 111: (see above)

src/IO/output_LTO_perf.jl
argonaut22: perhaps a better var name is `correlation_type`, since that is the extent of its influence

src/mission/odperformance.jl
```diff
@@ -1,352 +1,478 @@
...
-        desTAS[i] = TASdes/kts_to_mps # Save descent TAS separately, convert to knots
-        M0s[i] = Mach
-        Reunits[i] = V*ρ/μ
+    MNcr = ac.parad[iaMach, ipcruise1]
+    Tt41max = maximum(ac.pared[ieTt41, :])  #TODO set to max temp
+    Tt41CR = ac.pared[ieTt41,ipcruise1]
+    println(@sprintf("%12s %12s %12s %12s %12s %12s %12s %12s %12s %12s %12s %12s %12s %12s", 
+    "FL", "TAS", "CAS", "Mach", "Fn", "L/D", "Tt4max", "Tmetmax", "FFmax", "Tt4cruise", "Tmetcruise", "FFcruise", "CL", "CLh"))
+    # integrate trajectory over climb
+    for   i = 1:N
```
argonaut22, line 71: labelled indices are nice
```suggestion
    for   iFL = 1:N
```

src/mission/odperformance.jl
```diff
@@ -507,17 +632,139 @@ function TAS_CAS(TAS, h)
...
+        fig, ax = plt.subplots()
+    end
+    ffnorm = ff'./minimum.(eachcol(ff))
+
+    ax.plot(ffnorm', H)
+
+end
+
+
+## FOR 77W__
```
argonaut22, line 647: what in tarnation is 77W

argonaut22 (commented):
mostly best practices suggestions and continued distaste for `aircraft_type`

may i also recommend a description in the IO part of the docs?

PR #130, opened 2025-02-13
Adding variable polytropic efficiencies

ngomezve:
This PR adds support for variable compressor polytropic efficiencies. In the current version, the efficiencies are artificially set to the maximum ones at all design points. The parameters are based on those in the original TASOPT (compressor version ccc), with some modifications to the high-pressure compressor to better match a realistic map.

A small change is done to the engine initialization flags. Before, the option `initializes_engine = false` would initialize the engine in the call to `tfcalc`. This is confusing so the logic has been inverted: now `initializes_engine = true` initializes the engine.


ngomezve:
Map comparison between current map models and E3 data (courtesy of @WyattGiroux). E3 data in grey.

image
image

ngomezve:
The ccc option is how the FORTRAN version referred to one of the maps. Basically a comment next to the maps that said "ccc".

WyattGiroux:
Added links to the journal paper and NASA reports for the E3 compressors to both the comments in the code and the docs at the bottom of the Propulsion systems/Propulsion system page (includes table of curve fit parameters and plot comparing E3 and Drela fit for each compressor).

@argonaut22 or @aditeyashukla could you review and approve if everything looks good?

Project.toml
```diff
@@ -11,6 +11,7 @@ DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
 Dierckx = "39dd38d3-220a-591b-8e3c-4c3a8c710a94"
 DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
askprash, line 14:
```suggestion
```

Project.toml
```diff
@@ -27,4 +28,5 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [compat]
+Documenter = "1.8.1"
```
askprash, line 31:
```suggestion
```

PR #134, opened 2025-02-20
Removing overconstraints in off-design aircraft

ngomezve:
When TASOPT aircraft were analyzed off-design in `fly_off_design!`, the user was free to independently set the Mach number, altitude, and lift coefficient at cruise. In general, this leads to $L \neq W$. This PR changes `fly_off_design!` so that the cruise altitude is calculated for a desired lift coefficient.


example/cryo_input.toml
askprash, line 1: Was this meant to be added to this PR or an accidental commit?

ngomezve: The docs break without this because I added Roots to .atmosphere for the density-altitude calculation. So this fix is part of the PR.

src/mission/off_design.jl
```diff
@@ -238,7 +235,22 @@ function fly_off_design!(ac, mi = 1; itermax = 35, initializes_engine = true)
...
-    set_ambient_conditions!(ac, ipcruise1)
+    #Calculate start-of-cruise altitude from desired lift coefficient
+    # Use cabin volume to get buoyancy weight
+    ρcab = max(parg[igpcabin], pare[iep0, ipcruise1]) / (RSL * TSL)
+    WbuoyCR = (ρcab - pare[ierho0, ipcruise1]) * gee * parg[igcabVol]
+
+    ip = ipcruise1
+    We = WMTO * para[iafracW, ip]
+    CL = para[iaCL, ip]
+    u0 = pare[ieu0, ip]
```
askprash, line 246: Are we better of setting the cruise Mach number and then finding the right pressure, esp cause it looks like further down when the `set_ambient_conditions!` is called the flight speed is overwritten based on the fixed cruise Mach number. I think in typical missions this will mostly not be an issue since $a$ is going to be practically constant if all missions end up in the tropopause, but I could see this potentially being an issue if the off-design mission is very different. Maybe more robust to just use the Mach number directly here and recast the lift balance as follows:
$$L = \frac{\gamma}{2}pM^2C_LS$$,
so for fixed M, and $C_L$ you effectively get $p \propto L$.

ngomezve: The speed gets iterated in the weight loop so eventually you converge to the right combination of speed and altitude.

example/cryo_input.toml
ngomezve: Sorry, incorrect comment. The changes to that input file are to simplify the warnings for the test outputs. Not really relevant to this PR but something nice to add.

src/mission/off_design.jl
askprash: That makes sense - because it's being iterated the velocity will eventually converge to the right one even if it starts of being quite a bit different from a quick test. The off-design mission needs to be quite a bit different for the methods to be different. We can revisit this later if we have new features like constraining the altitude for short missions where it may not make sense to climb up to optimal CL.

askprash (approved):
Looks good to me!

PR #140, opened 2025-02-23
Replace `pari` with new `ac.Options` struct

argonaut22:
In addition, replaces the integer flag style with the new guidelines:
- for `Bool`s (i.e., `<verb>s_<subject>`) 
- `String`s (i.e., `opt_<descriptive>`)

Supersedes #98 

I didn't realize how many files this PR would touch. God help us.


src/IO/default_input.toml
```diff
@@ -80,12 +78,12 @@
     inner_vessel_support_angle = "80 deg"
     
     #Insulation options: "rohacell41S", "polyurethane27", "polyurethane32", "polyurethane35", "vacuum", "mylar", "microspheres"
-    size_insulation = true #Flag to decide whether to size insulation for boiloff
+    sizes_insulation = true #Flag to decide whether to size insulation for boiloff
```
askprash, line 81: I have a preference for the imperative clause for these boolean flags rather than descriptive - "size_insulation" "initialize_x". But I have a stronger preference to get rid of the pari array

src/IO/default_input.toml
```diff
@@ -226,9 +224,9 @@
     BLI_frac = 0.0 #Fraction of fuselage BL KE defect ingested
 
 [Wing]
-    wing_planform = 1 # 0: Wing cantilever, plain
-                      # 1: Wing cantilever with engine
-    strut_braced_wing = false
+    has_engine = true # false: Wing cantilever, plain
```
askprash, line 227: Does this input get overridden? In read_intput it looks like this variable is set by comparing `engloc`. This option might no longer be needed/ is a duplicate of `engloc`

src/data_structs/wing.jl
```diff
@@ -128,20 +127,20 @@ $TYPEDFIELDS
...
     CL_max_fwd_CG::Float64 = 0
     """Tail Minimum static margin"""
     SM_min::Float64 = 0
     """Max Tail down load. Tail download param at max load case"""
     CL_CLmax::Float64 = 0
     """Number of Tails"""
     ntails::Float64 = 0
-    """Move wingbox factor. 0="fix" wing position ,1=move wing to get CLh="CLhspec" in cruise, 2= move wing to get min static margin = "SMmin"  """
-    move_wingbox::Int64 = 0
+    """Move wingbox selection for longitudinal stability analysis. "fixed" = static wing position ,"fixed_CLh" move wing to get CLh="CLhspec" in cruise, "min_static_margin" = move wing to get min static margin = "SMmin"  """
```
askprash, line 142: This is an interesting one, should this and the above be in the aircraft level `options` struct rather than the wing structure? Yes, the wing moves based on this parameter but the choice of which parameter to design around affects the wing and tail for example.

src/data_structs/aircraft.jl
```diff
@@ -11,29 +11,29 @@ Overloads Base.summary to print a summary of the `aircraft` model.
...
 - `para::AbstractArray{Float64}` : Aero parameters                       
 - `pare::AbstractArray{Float64}` : Engine parameters 
-- `sized::AbstractVector{1,Bool}`: flag if aircraft is sized (default is `[false]`)
+- `is_sized::AbstractVector{1,Bool}`: flag if aircraft is sized (default is `[false]`)
 
-For devs: the indices for accessing specific data are defined in `/src/misc/index.inc`. Refer to the sample input file (`/src/IO/default_input.toml` and `read_input.jl`) for usage.
+For devs: the indices for accessing specific data are defined in `/src/data_structs/index.inc`. Refer to the sample input file (`/src/IO/default_input.toml` and `read_input.jl`) for usage.
 """
-Base.@kwdef mutable struct aircraft #inner constructor
+@kwdef mutable struct aircraft #inner constructor
```
askprash, line 23: What does the "inner constructor" mean here?

argonaut22: not sure if this was me, but I think it's meant to indicate that @kwdef is creating constructors (though a cursory search says they're outer, not inner)

argonaut22: I don't think it matters and the comment can get binned

src/IO/default_input.toml
argonaut22: 😬
image

if it's all the same, leaving it as-is is my preference

src/data_structs/wing.jl
argonaut22: Yeah, I thought about this as well and I could see it. 

Ultimately went with a push-anything-that's-not-bolted-down-to-submodels approach. I figured it made sense to put the option near the relevant parameters. Still, either works.

src/IO/default_input.toml
argonaut22: Correct that it's dependent on `engloc`.
- For simplicity, I think binning one of the vars is easiest
- However, looking ahead, is there a case for keeping engine location non-binary? Having a `has_engine` toggle on fuselage and wing sounds good, and engloc can parse from that perhaps? For now, having both would have to error as unsupported

argonaut22: Second thought, no reason we can't do that from an `engloc` input.

src/IO/default_input.toml
ngomezve: I also have a mild preference for the imperative - it feels more natural. But also willing to accept this descriptive flag option.

argonaut22:
For anyone wondering, performance is comparable, adding a few ms to the benchmarking cases:

This branch
image


Upstream/main
image

src/data_structs/wing.jl
argonaut22: irl discussion: anything that touches more than one subsystem should be in `aircraft.options`

src/IO/default_input.toml
argonaut22: i stand by the useful distinction:
- `size_aircraft` is a function thus imperative
- `sizes_aircraft` is a flag

argonaut22:
updated and ready to merge after all the upstream PRs. All subsequent edits will go in #141

src/cryo_tank/tanktools.jl
```diff
@@ -1,24 +1,22 @@
...
     - `parg::Vector{Float64}`: vector with aircraft geometric parameters
     - `para::Array{Float64}`: array with aircraft aerodynamic parameters
     - `pare::Array{Float64}`: array with aircraft engine parameters
     
     **Outputs:**
     - `t::mdot`: fuel mass flow rate out of the tank (kg/s)
 """
-function find_mdot_time(t::Float64, pari::Vector{Int64}, parg::Vector{Float64}, para::Array{Float64}, pare::Array{Float64})
-    nftanks = pari[iinftanks]
+function find_mdot_time(t::Float64, fuselage_fueltank_count::Int64, parg::Vector{Float64}, para::Array{Float64}, pare::Array{Float64})
```
askprash, line 16:
```suggestion
function find_mdot_time(t::Float64, fueltank_count::Int64, parg::Vector{Float64}, para::Array{Float64}, pare::Array{Float64})
```
Probs can simplify by dropping "fuselage" since rn we are not going to have situations where we have multiple non-fuselage fuel tanks (which for now is just the wing)

docs/make.jl
```diff
@@ -50,13 +50,13 @@ makedocs(
...
         "data_io/data_basics.md",
+        "data_io/structs.md",
         "data_io/data_io.md"
         ],
 
     "Miscellaneous" => Any[
-        "misc/structs.md",
         "misc/dreladocs.md",
-        "misc/misc.md",
+        # "misc/misc.md",
```
askprash, line 59:
```suggestion
```

docs/src/balance/balance.md
```diff
@@ -2,8 +2,12 @@
...
+- [`balance()`](@ref TASOPT.balance) makes adjustments as described below to achieve pitch trim. It calculates the aircraft's resulting **center of gravity (`xCG`)**, **center of pressure (`xCP`)**, and **neutral point (`xNP`)** at a specific flight point. To meet the pitch trim requirement, the routine adjusts one of (i) the horizontal tail's lift coefficient, (ii) its area, or (iii) the axial location of the wing box. In *almost all* the use cases, only option (i) adjusting the horizontal tail's lift coefficient is relevant for trim calculations.
+
+!!! warning "Some `balance()` options change the aircraft design"
+    Exercise caution if options (ii) or (iii) are being used as these options *alter the design of the aircraft* (i.e., they should not be used in off-design cases).
+
+- [`size_htail()`](@ref TASOPT.size_htail) performs a more involved stability analysis: it determines the **horizontal tail area (`Sh`)** and **wing box position (`xwbox`)**, solving for pitch trim and static stability across all flight conditions.
```
askprash, line 10:
```suggestion
- [`size_htail()`](@ref TASOPT.size_htail) simultaneously sizes **horizontal tail area (`Sh`)** and positions the **wing box (`xwbox`)**, by solving for pitch trim and static stability across all flight conditions.
```
A little much to call this an involved stability analysis.

askprash (changes requested):
Review part 1.

docs/src/balance/balance.md
argonaut22: the siren's song of adverbs
agreed

src/aero/fusebl.jl
```diff
@@ -93,7 +93,7 @@ function fusebl!(fuse, parm, para, ip)
       nc = 30
 
       nbl, iblte =  axisol!(xnose,xend,xblend1,xblend2,Sfuse, 
-                            anose,btail,ifclose,
+                            anose,btail,Int(tapers_to_edge),
```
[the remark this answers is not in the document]
argonaut22: from before I decided to carry bool farther down the layers. good catch

src/balance/balance.jl
```diff
@@ -308,18 +302,23 @@ The routine considers:
...
-      - If `ixwmove == 0`, the wing location is fixed (no movement).
-      - If `ixwmove == 1`, `xwbox` is adjusted to satisfy cruise tail lift conditions.
-      - If `ixwmove == 2`, `xwbox` is adjusted to enforce static margin at aft-CG.
-      
-Note the two flags can be set independently and affect how the two stability residuals are driven to zero. The 2x2 system is built sequentially as annotated in the source code for this function. 
+**Note**: two flags determine the sizing strategy:
+
+      - `htail.opt_sizing`
+            * = "fixed_Vh"      adjust Sh    (horizontal tail area)
+            * = "max_fwd_CG"    adjust CLh   (horizontal tail cl)
```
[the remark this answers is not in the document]
argonaut22: ohh. there is a misunderstanding here. will clarify

src/balance/balance.jl
```diff
@@ -656,7 +653,7 @@ function htsize(ac, paraF, paraB, paraC)
       dxeng2wbox = parg[igdxeng2wbox]
       parg[igxeng] = xwbox - dxeng2wbox
 
-      if (iHTsize == 1)
+      if compare_strings(ac.htail.opt_sizing, "fixed_Vh")
```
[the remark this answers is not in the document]
argonaut22: read_input could compare the strings once and set a function field in the way that generalized engine now does

src/cryo_tank/tanktools.jl
argonaut22: that's true. that said, i had the thought at some point "does nftanks include the wing? are we counting the wingbox as separate?" silly when you know how it works, less so before

argonaut22: that said, this function rn is only applicable to fuse tanks. done for brevity

askprash (approved):

Issue #125, opened 2025-02-10
`read_externals.jl` don't work with new structs


argonaut22:
...and we may not fix it (and by it, i mean `read_mdl` and `read_jl`). 

This flew under the radar of the latest IO updates.

Since we can't trivially convert generic `par`s from previous versions to the new way of doing things, i don't think it's worth the time to retain backwards-compatibility with those savefiles. Thus, i think these functions should be deleted unless anyone can think of a use case to keep them.

argonaut22:
Also in this boat: `load_aircraft()` in `LoadModel.jl`

PR #141, opened 2025-02-23
JOSS editorial PR

argonaut22:
In which we take steps to make this thing more accessible. Including:
- [x] removing unused functions, files
- [x] reorganizing the tree to reduce redundancy/confusion
- [x] renaming functions for clarity
- [x] add warnings for functionalities that aren't currently supported but that we retain code for (for the future; see: turboelectrics); note, we _will_ remove `parpt` and co since we won't be developing for `par` arrays
- [x] a pass at updating docstrings and documentation to reflect the `struct` updates

Should also address: #125 

NOTE: this branch will need a `rebase` once #140 gets pulled in to remove the duplicate commits


argonaut22:
Renaming changes summarized here:

image

src/mission/mission_iter.jl
askprash, line 1: Given the renaming done for `fly_offdesign_mission` --> `fly_mission`, this might be a function we make "internal" by calling it `_mission_iterations` for example.

src/aero/wing_calculations.jl
askprash, line 1: The symmetry of having `drag.jl`, `wing_drag.jl`, `fuselage_drag.jl` might be nice and keep the long drag calculations separate from lift and CM calculations in this file.

askprash: Perhaps also`trefftz` --> `induced_drag.jl`?

askprash (approved):

PR #151, opened 2025-03-17
Improvements to cabin sizing

ngomezve:
This PR adds minor enhancements to the functions that can be used to resize the aircraft fuselage. In particular, three issues are addressed. 
- Magic numbers are eliminated from the source code: these were default parameters that were used even when a different one was specified. For example, the seats were plotted in the stickfig diagrams with the default pitch and width, despite these parameters already being inputs.
- The longitudinal offsets before and after the seats can now be explicitly specified. Before, a 10 ft offset before the seats and a 0 or 5 ft (if there were tanks) offset after the seats were used.
- The cargo bay when the aircraft is a double decker is sized to carry a given unit load device (ULD). Before, there was only a minimum height constraint and the width was not checked at all. The new constraint ensures that the ULD can fit in the cargo bay. A dictionary with the sizes of standard ULDs is introduced.


aditeyashukla (approved):
Looks good! Very well organized

PR #164, opened 2025-05-05
Improving and speeding up induced drag calculations

askprash:
- [ ] Create a "wake system" that is only geometry related
- [ ] Propagate changes into induced drag calculations
- [ ] add logic to only update wake system when wake geometry changes otherwise just calculate drag by matrix multiplication

### DIFF
--- a/src/aero/WakeGeometry.jl
+++ b/src/aero/WakeGeometry.jl
@@ -1,0 +1,19 @@
+
+using StaticArrays
+using LinearAlgebra
+using DocStringExtensions
+
+"""
+    $(TYPEDEF)
+
+Two-dimensional point type using StaticArrays for stack allocation.
+First component is y-coordinate, second is z-coordinate.
+# Examples
+```julia-repl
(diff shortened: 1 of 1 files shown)


src/aero/WakeGeometry.jl
```diff
@@ -0,0 +1,268 @@
...
+    length::Float64
+    "y-extent of element"
+    Δy::Float64
+    "z-extent of element"
+    Δz::Float64
+    "Unit normal vector to the element"
+    unit_normal::Point2D
+    
+    # Inner constructor to help in calculating the control point and the normal
+    function WakeElement(p1::Point2D, p2::Point2D;
```
ian-ross: Can't you say:

```julia
function WakeElement(p1::Point2D, p2::Point2D; control_point::Point2D=(p1 + p2) * 0.5)
```

to avoid the `nothing` checks? (Based on https://docs.julialang.org/en/v1/manual/functions/#Optional-Arguments and https://docs.julialang.org/en/v1/manual/functions/#Evaluation-Scope-of-Default-Values)

Issue #165, opened 2025-05-05
Off-design fuselage drag requires cruise altitude


ngomezve:
In the current off-design framework, the off-design lift coefficient is assumed to be an input (which can then be optimized) and the corresponding cruise altitude is subsequently calculated so that $L=W$. However, the fuselage drag areas are calculated outside of the weight loop, in which the altitude is recalculated. This leads to the fuselage drag areas not being computed with the right altitude conditions. 

For example, if one runs an off-design case twice by calling `fly_mission!()`, the resulting fuel demand will be different in the second case compared to the first one due to the fuselage drag area issue.

There are two possible solutions: 
1. Specify the flight altitude and calculate the corresponding lift coefficient. This would work but the user may not know what a good guess for the initial altitude is a priori, since it is usually better to fly at a lift coefficient near the design one.
2. Recalculate the fuselage drag areas in the weight loop, either every iteration or every few iterations. This would keep the same off-design framework but it adds an overhead since the fuselage drag calculation is expensive.

ngomezve:
Fixed in PR #166

PR #166, opened 2025-05-05
Fixing issue with off-design fuselage drag

ngomezve:
This PR addresses issue #165. The off-design fuselage drag is calculated outside of the weight loop; however, this requires knowledge of the cruise altitude, which is modified in the weight loop to keep a desired lift coefficient. 

This PR modifies the off-design function so that the user can specify either the cruise altitude or the lift coefficient, with the other parameter being then calculated. The altitude is assumed to be specified by default as this is most efficient.

This PR also fixes a small bug with the maximum payload in the payload-range diagram plots.


ngomezve:
Some benchmarks. 

With the current PR, running `fly_mission!(ac,2)` for the default input `ac`:
BenchmarkTools.Trial: 120 samples with 1 evaluation per sample.
 Range (min … max):  34.148 ms … 73.243 ms  ┊ GC (min … max): 0.00% … 47.12%
 Time  (median):     39.248 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   41.608 ms ±  7.131 ms  ┊ GC (mean ± σ):  6.14% ±  9.49%

  ▁▂▇▂█ ▅ ▁▅▄       ▁▂
  ███████████▃▃▃▅▃▆▃██▅█▃▅▃▅▁▃▅▁▁▃▁▅▁▁▃▃▁▃▁▁▁▁▅▃▁▁▁▁▁▁▁▁▃▁▁▁▃ ▃
  34.1 ms         Histogram: frequency by time        66.4 ms <

 Memory estimate: 22.43 MiB, allocs estimate: 583396.

Keeping the lift coefficient as the input but recalculating the fuselage drag:
BenchmarkTools.Trial: 77 samples with 1 evaluation per sample.
 Range (min … max):  55.697 ms … 145.367 ms  ┊ GC (min … max): 0.00% … 17.12%        
 Time  (median):     61.240 ms               ┊ GC (median):    6.14%
 Time  (mean ± σ):   65.474 ms ±  14.913 ms  ┊ GC (mean ± σ):  7.00% ±  4.85%        

    ▁▄█▃
  ▅▅████▇▁▇▆▃▅▃▁▄▁▁▁▃▁▁▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▁▁▃▁▁▃ ▁
  55.7 ms         Histogram: frequency by time          123 ms <

 Memory estimate: 49.05 MiB, allocs estimate: 1135521.

Same, but recalculating the fuselage drag every 5th iteration:
BenchmarkTools.Trial: 113 samples with 1 evaluation per sample.
 Range (min … max):  37.545 ms … 75.558 ms  ┊ GC (min … max): 0.00% … 41.34%
 Time  (median):     43.389 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   44.501 ms ±  5.534 ms  ┊ GC (mean ± σ):  7.03% ±  8.48%

    ▂▄█ ██▂▂  ▄ ▂  ▄▆ ▆  ▆   ▂
  █▆█████████▆█▄█▁███▆█▄██▄▄▄█▄▄▆▄▄▄▄▄▁▁▁▄▁▁▁▁▄▁▁▁▁▁▁▁▁▁▁▁▁▁▄ ▄
  37.5 ms         Histogram: frequency by time        63.6 ms <

 Memory estimate: 26.90 MiB, allocs estimate: 676983.

src/mission/fly_mission.jl
```diff
@@ -304,4 +280,56 @@ function fly_mission!(ac, imission = 1; itermax = 35, initializes_engine = true)
...
+    ρcab = max(parg[igpcabin], pare[iep0, ipcruise1]) / (RSL * TSL)
+    WbuoyCR = (ρcab - ρ0) * gee * parg[igcabVol]
+    
+    ip = ipcruise1
+    We = WMTO * para[iafracW, ip]
+    u0 = pare[ieu0, ip]
+    BW = We + WbuoyCR # Weight including buoyancy
+    S = wing.layout.S
+
+    if compare_strings(specifying_cruise, "altitude")
```
argonaut22, line 321: A good preemptive error catch might be:

```suggestion
    if compare_strings(specifying_cruise, "altitude")   
        if para[iaCL, ipcruise,1] != para[iaalt, ipcruise, imission] #if CLs are prescribed but 
            @warn "An off-design CL is specified, but the analysis is set to rewrite it. Please review your inputs, especially `specifying_cruise`, if this is not the desired behavior"
        end
```

along with a similar one for the following `elseif`

src/IO/output_plots.jl
```diff
@@ -973,12 +973,13 @@ Function to plot a payload range diagram for an aircraft
...
     - `itermax::Int64`: Max Iterations for fly_mission! loop (Optional).
     - `initializes_engine::Boolean`: Use design case as initial guess for engine state if true (Optional)
+    - `specifying_cruise::String`: option for whether cruise altitude or lift coefficient is specified. Options are "altitude" or "lift_coefficient"
     - `Ldebug::Bool`: verbosity flag. false by default, hiding outputs as PR sweeps progress (Optional).
 """
 function PayloadRange(ac_og::TASOPT.aircraft; 
     Rpts::Integer = 20, Ppts::Integer = 20, OEW::Bool = false,
     filename::String = "", 
-    itermax::Int64 = 35, initializes_engine::Bool = true,
+    itermax::Int64 = 35, initializes_engine::Bool = true, specifying_cruise = "lift_coefficient",
```
argonaut22, line 982: May I suggest something like `specifying_cruise` -> `opt_prescribed_cruise_parameter`, to follow the partially-realized and yet-to-be-formalized best practices, ofc propagating to the docstrings, other fxns, etc.

src/mission/fly_mission.jl
```diff
@@ -9,12 +9,13 @@ Runs the aircraft through the specified mission, computing and converging the fu
 - `imission::Int64`: Off design mission to run (Default: 1)
 - `itermax::Int64`: Maximum iterations for sizing loop
 - `initializes_engine::Boolean`: Use design case as initial guess for engine state if true
-
+- `specifying_cruise::String`: option for whether cruise altitude or lift coefficient is specified. Options are "altitude" or "lift_coefficient"
```
argonaut22, line 12: see above for comment on param name pls

src/IO/output_plots.jl
argonaut22: also, you could probably get away with "CL" instead of "lift_coefficient" but I'm fine either way

argonaut22 (changes requested):
A few naming/docs points but overall looks good. 

It would also be good to have a test show that both parameter specification cases are working as intended (probably sufficient to do it via `fly_mission`). Not just for unit testing, but to prove to ourselves that this is right

src/IO/output_plots.jl
ngomezve: I don't think that `opt_prescribed_cruise_parameter` is much better than `specifying_cruise` but also don't have a strong opinion against it, so will make the change. Will also go with `CL` as it is more compact.

argonaut22 (approved):

argonaut22 (approved):

PR #169, opened 2025-05-14
Speeding up tank calculations

ngomezve:
This PR introduces a significant speedup to the tank pressure calculation functions. Most of this is achieved by removing type instabilities that result in slower code. A further speed improvement is achieved in the fuselage heat transfer function by approximating air's $c_p$ as constant.

By defining the aircraft data arrays as `Array` instead of `AbstractArray`, the default aircraft is also sped up.

### DIFF
--- a/src/IO/read_input.jl
+++ b/src/IO/read_input.jl
@@ -464,6 +464,11 @@ if !(has_wing_fuel) #If fuel is stored in fuselage
     fuse_tank.qfac = readfuel_storage("heat_leak_factor")
     fuse_tank.pfac = readfuel_storage("pressure_rise_factor")
 
+    for (i,altTO) in enumerate(parm[imaltTO, :]/1e3)
+        T_std, _, _, _, _ = atmos(altTO)
+        push!(fuse_tank.TSLtank, parm[imT0TO,i] - T_std + Tref)
+    end
+    
     has_vacuum = TASOPT.CryoTank.check_vacuum(fuse_tank.material_insul) #flag to check if an outer vessel is needed
 
     if has_vacuum #If tank is double-walled
--- a/src/IO/read_input.jl
+++ b/src/IO/read_input.jl
@@ -462,7 +462,6 @@ if !(has_wing_fuel) #If fuel is stored in fuselage
     fuse_tank.ew = readfuel_storage("weld_efficiency")
     fuse_tank.ullage_frac = readfuel_storage("ullage_fraction")
     fuse_tank.qfac = readfuel_storage("heat_leak_factor")
-    fuse_tank.TSLtank = Temp(readfuel_storage("SL_temperature_for_tank"))
     fuse_tank.pfac = readfuel_storage("pressure_rise_factor")
 
     has_vacuum = TASOPT.CryoTank.check_vacuum(fuse_tank.material_insul) #flag to check if an outer vessel is needed
--- a/src/cryo_tank/pressure.jl
+++ b/src/cryo_tank/pressure.jl
@@ -19,13 +19,13 @@ end
 """
 Structure with inputs for tank pressure calculation.
 """
-struct tank_inputs
+struct tank_inputs{F1, F2, F3}
     """Heat rate as a function of time (W)"""
-    Q_calc::Function
+    Q_calc::F1
     """Work rate as a function of time (W)"""
-    W_calc::Function
+    W_calc::F2
     """Liquid mass flow rate as a function of time (kg/s)"""
-    mdot_calc::Function
+    mdot_calc::F3
 end
 
 """
--- a/src/cryo_tank/pressure.jl
+++ b/src/cryo_tank/pressure.jl
@@ -137,7 +137,7 @@ This function returns the time derivatives for pressure and liquid volume fill f
     **Outputs:**
     - `dydt::Vector{Float64}`: vector with the state derivatives in time
 """
-function TankDerivatives(t::Float64, y::Vector{Float64}, u::tank_inputs, params::tank_params)
+function TankDerivatives(t::Float64, y::AbstractVector{Float64}, u::tank_inputs, params::tank_params)
     #Extract states 
     p = y[1]
     β = y[2]
@@ -164,7 +164,6 @@ function TankDerivatives(t::Float64, y::Vector{Float64}, u::tank_inputs, params:
     end
 
     #Calculate derivatives
-    dydt = zeros(6)
     if p < pmax
         mdot_vent = 0.0
     else
@@ -177,13 +176,17 @@ function TankDerivatives(t::Float64, y::Vector{Float64}, u::tank_inputs, params:
     #Total mass flow rate for beta
     mdot_tot = mdot + mdot_vent
 
+    dp_dt = dpdt(mix_current, Q, W, mdot, xout, mdot_vent, xvent, V, α)
+    dβ_dt = dβdt(mix_current, dp_dt, mdot_tot, V) #dβ/dt
     #Calculate derivates and store them
-    dydt[1] = dpdt(mix_current, Q, W, mdot, xout, mdot_vent, xvent, V, α) #dp/dt
-    dydt[2] = dβdt(mix_current, dydt[1], mdot_tot, V) #dβ/dt
-    dydt[3] = -mdot_tot #dM_tank/dt
-    dydt[4] = mdot #dMburn/dt
-    dydt[5] = mdot_vent #dMvent/dt
-    dydt[6] = mdot_boiloff(mix_current, dydt[2], dydt[1], mdot_liq, V) #dMboiloff/dt
+    dydt = @SVector[
+        dp_dt, 
+        dβ_dt,
+        -mdot_tot, #dM_tank/dt
+        mdot, #dMburn/dt
+        mdot_vent, #dMvent/dt
+        mdot_boiloff(mix_current, dβ_dt, dp_dt, mdot_liq, V) #dMboiloff/dt
+    ]
     return dydt
 end
 
--- a/src/cryo_tank/pressure.jl
+++ b/src/cryo_tank/pressure.jl
@@ -178,7 +178,7 @@ function TankDerivatives(t::Float64, y::AbstractVector{Float64}, u::tank_inputs,
 
     dp_dt = dpdt(mix_current, Q, W, mdot, xout, mdot_vent, xvent, V, α)
     dβ_dt = dβdt(mix_current, dp_dt, mdot_tot, V) #dβ/dt
-    #Calculate derivates and store them
+    #Calculate derivatives and store them
     dydt = @SVector[
         dp_dt, 
         dβ_dt,
--- a/src/cryo_tank/tankWmech.jl
+++ b/src/cryo_tank/tankWmech.jl
@@ -7,7 +7,7 @@
       **Inputs:**
       - `fuse::Fuselage`: fuselage object.
       - `fuse_tank::fuselage_tank`: fuselage tank object.
-      - `t_cond::Float64`: Vector with tank isulation layer thickness. Provided separately from fuse_tank as it changes during 
+      - `t_cond::Vector{Float64}`: Vector with tank isulation layer thickness. Provided separately from fuse_tank as it changes during 
       non-linear solve process.
 
       **Outputs:**
@@ -35,7 +35,7 @@ function size_inner_tank(fuse::Fuselage, fuse_tank::fuselage_tank, t_cond::Vecto
       #---------------------------------
       # Unpack parameters in fuse_tank
       #---------------------------------
-      Rfuse = fuse.layout.radius
+      Rfuse = structures.Rfuse(fuse)
       fuse_cs = fuse.layout.cross_section #Fuselage cross section
 
       Wfuel = fuse_tank.Wfuelintank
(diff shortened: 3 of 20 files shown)


ngomezve:
main
---------------------------------------
DEFAULT aircraft
BenchmarkTools.Trial: 5 samples with 1 evaluation per sample.
 Range (min … max):  44.036 ms … 49.496 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     45.228 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   45.852 ms ±  2.105 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █          ██ █                                           █  
  █▁▁▁▁▁▁▁▁▁▁██▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  44 ms           Histogram: frequency by time        49.5 ms <

 Memory estimate: 25.83 MiB, allocs estimate: 690761.
HYDROGEN aircraft
BenchmarkTools.Trial: 5 samples with 1 evaluation per sample.
 Range (min … max):  355.218 ms … 388.735 ms  ┊ GC (min … max): 8.53% … 7.37%
 Time  (median):     375.189 ms               ┊ GC (median):    7.64%
 Time  (mean ± σ):   372.529 ms ±  12.256 ms  ┊ GC (mean ± σ):  7.41% ± 1.53%

  █                      █            ██                      █  
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁██▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  355 ms           Histogram: frequency by time          389 ms <

 Memory estimate: 366.69 MiB, allocs estimate: 8119595.

dev_tankspeedup
------------------
DEFAULT aircraft
BenchmarkTools.Trial: 5 samples with 1 evaluation per sample.
 Range (min … max):  38.089 ms … 42.137 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     38.723 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   39.408 ms ±  1.711 ms  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █        ▁                 ▁                              ▁  
  █▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  38.1 ms         Histogram: frequency by time        42.1 ms <

 Memory estimate: 23.33 MiB, allocs estimate: 547012.
HYDROGEN aircraft
BenchmarkTools.Trial: 5 samples with 1 evaluation per sample.
 Range (min … max):  297.480 ms … 421.711 ms  ┊ GC (min … max): 0.00% … 8.42%
 Time  (median):     316.266 ms               ┊ GC (median):    7.58%
 Time  (mean ± σ):   334.806 ms ±  49.887 ms  ┊ GC (mean ± σ):  6.41% ± 3.59%

  █     █  █     █                                            █  
  █▁▁▁▁▁█▁▁█▁▁▁▁▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  297 ms           Histogram: frequency by time          422 ms <

 Memory estimate: 286.13 MiB, allocs estimate: 6026788.

src/cryo_tank/pressure.jl
```diff
@@ -137,7 +137,7 @@ This function returns the time derivatives for pressure and liquid volume fill f
     **Outputs:**
     - `dydt::Vector{Float64}`: vector with the state derivatives in time
 """
-function TankDerivatives(t::Float64, y::Vector{Float64}, u::tank_inputs, params::tank_params)
+function TankDerivatives(t::Float64, y::AbstractVector{Float64}, u::tank_inputs, params::tank_params)
```
askprash, line 140: Have you tried this one with dydt being a sized vector (or @MVector) _outside the function definition_  and passing in `dydt` to the function instead of the output? Since you know the size will always be 6. I'm curious if dydt preallocated outside will be even faster than the SVectors being created each time.

src/cryo_tank/tankWmech.jl
```diff
@@ -119,11 +119,11 @@ function size_inner_tank(fuse::Fuselage, fuse_tank::fuselage_tank, t_cond::Vecto
       # Insulation weight
       N = length(t_cond) #Number of insulation layers
       #Initialize storage vectors
-      Vcyl_insul = zeros(N)
-      Winsul = zeros(N)
-      Shead_insul = zeros(N + 1) #add one for first (tank wall) surface 
-      Vhead_insul = zeros(N)
-      rho_insul = zeros(N)
+      Vcyl_insul = zeros(Float64, N)
```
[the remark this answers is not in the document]
askprash: Well actually, do we need all the individual layer `Shead_insul`, `rho_insul`, `Vhead_insul` to be saved and stored? If we don't need it we could just get rid of the arrays and reuse a work variable and only store an array of weights or even the total weight?

src/data_structs/layout.jl
```diff
@@ -87,6 +87,9 @@ function Base.getproperty(obj::MultiBubble, sym::Symbol)
     end
 end  # function Base.getproperty
 
+Rfuse(cs::SingleBubble) = cs.radius
+Rfuse(cs::MultiBubble) = cs.radius
```
askprash, line 91: Not quite sure why we need this but even if we do we don't need to go define this for every type. If the structure has a radius it will return a radius
```suggestion
Rfuse(cs) = cs.radius 
```

src/data_structs/fuselage.jl
```diff
@@ -71,6 +71,8 @@ function dx_cabin(fuse::Fuselage)
     return fuse.layout.x_end_cylinder - fuse.layout.x_start_cylinder
 end
 
+Rfuse(fuse::Fuselage) = Rfuse(fuse.layout.cross_section)
```
askprash, line 74:
```suggestion
Rfuse(fuse::Fuselage) = fuse.layout.cross_section.radius
```

src/cryo_tank/tankWmech.jl
```diff
@@ -35,7 +35,7 @@ function size_inner_tank(fuse::Fuselage, fuse_tank::fuselage_tank, t_cond::Vecto
       #---------------------------------
       # Unpack parameters in fuse_tank
       #---------------------------------
-      Rfuse = fuse.layout.radius
+      Rfuse = structures.Rfuse(fuse)
```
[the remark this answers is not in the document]
ngomezve: It made a world of a difference from the type stability standpoint. The key was that layout is an abstract type so the compiler does not know that it is a Float64. By defining this function, the type stability issue is resolved. I agree that it is cumbersome and counterintuitive.

src/data_structs/fuselage.jl
ngomezve: Same, I do this for stability.

src/cryo_tank/tankWmech.jl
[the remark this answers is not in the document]
ngomezve: So by creating the functions, the allocations are reduced by ~3%.

src/cryo_tank/pressure.jl
ngomezve: The [DifferentialEquations.jl](https://github.com/SciML/DifferentialEquations.jl) package needs the function to return the derivatives. I'm not sure why, but if I initialize the `dydt` vector outside the integral and then pass it to the derivative function, mutate it, and return it, the integrator no longer behaves as expected.

src/cryo_tank/tankWmech.jl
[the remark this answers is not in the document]
ngomezve: Good suggestion. These variables used to be returned but they are not anymore. No reason why we can't just overwrite them. The only exception is `Shead_insul` which does get returned and stored as you need it for the thermal calculations.

src/data_structs/fuselage.jl
ngomezve: Just checked that the stability issues in `size_inner_tank!` come back if this is used.

src/IO/read_input.jl
```diff
@@ -469,9 +469,13 @@ if !(has_wing_fuel) #If fuel is stored in fuselage
     fuse_tank.ew = readfuel_storage("weld_efficiency")
     fuse_tank.ullage_frac = readfuel_storage("ullage_fraction")
     fuse_tank.qfac = readfuel_storage("heat_leak_factor")
-    fuse_tank.TSLtank = Temp(readfuel_storage("SL_temperature_for_tank"))
     fuse_tank.pfac = readfuel_storage("pressure_rise_factor")
 
+    for (i,altTO) in enumerate(parm[imaltTO, :]/1e3)
```
argonaut22, line 474: a comment stating what's going on in this block pls

src/cryo_tank/tankWthermal.jl
```diff
@@ -21,25 +62,20 @@ for a given insulation thickness
 
 See [here](@ref fueltanks).
```
argonaut22, line 59:
```suggestion
See [cryo tanks docs](@ref fueltanks).
```

argonaut22: pls copy paste to the couple of other spots. github won't let me make the comments there for reasons unknown

src/cryo_tank/tanktools.jl
```diff
@@ -157,19 +160,19 @@ function find_Q_time(t::Float64, fuse::Fuselage, fuse_tank::fuselage_tank, fuelt
...
         end
     end
     return Q
 end
 
 """
-    analyze_TASOPT_tank(ac_orig, t_hold_orig::Float64 = 0.0, t_hold_dest::Float64 = 0.0, N::Int64 = 50)
+    analyze_TASOPT_tank(ac, t_hold_orig::Float64 = 0.0, t_hold_dest::Float64 = 0.0, N::Int64 = 50)
 
 This function analyses the evolution in time of a cryogenic tank inside a TASOPT aircraft model.
```
argonaut22, line 172:
```suggestion
This function analyses the evolution in time of a cryogenic tank inside a TASOPT aircraft model, accounting for operational constraints (i.e., holding times and venting).
```

is that about right^?

src/cryo_tank/tanktools.jl
```diff
@@ -157,19 +160,19 @@ function find_Q_time(t::Float64, fuse::Fuselage, fuse_tank::fuselage_tank, fuelt
...
 !!! details "🔃 Inputs and Outputs"
     **Inputs:**
-    - `ac_orig::aicraft`: TASOPT aircraft model
+    - `ac::aicraft`: TASOPT aircraft model
```
argonaut22, line 175:
```suggestion
    - `ac::aircraft`: TASOPT aircraft model
```

src/cryo_tank/tanktools.jl
```diff
@@ -246,18 +250,18 @@ function analyze_TASOPT_tank(ac_orig::aircraft, t_hold_orig::Float64 = 0.0, t_ho
...
-    y0 = [p0, β0, M0, 0.0, 0.0, 0.0] #Initial states
+    y0 = @SVector Float64[p0, β0, M0, 0.0, 0.0, 0.0] #Initial states
 
     tspan = (0.0, para_alt[iatime,end]) #start and end times
 
     dy_dt(y, p, t) = TankDerivatives(t, y, p[1], p[2]) #State derivatives
     #ODE problem; specify changes in mission segments for better speed
     prob = ODEProblem(dy_dt, y0, tspan, ODEparams, tstops = para_alt[iatime,:], reltol = 1e-8)
-    sol = solve(prob, Tsit5()) #Solve ODE problem using recommended solver
+    sol = solve(prob, Tsit5())::ODESolution #Solve ODE problem using recommended solver
```
argonaut22, line 260: how do we feel about a non-convergence warning? or a try-catch if that's appropriate?

Re, the warning, yes the user could likely piece it together via error tracing, but it'd be faster to point out this and perhaps some relevant troubleshooting quantities. Though I don't know this equation system or the sizing logic, errors seem plausible if not inevitable.

argonaut22 (commented):
Looks good overall. I mostly reviewed for clarity/comments and am trusting @askprash 's review of the speedup (last call for concerns there). 

Only other note: would like to see ~"cp is assumed constant, and can be changed" wherever is most relevant in the docs and docstrings

src/IO/read_input.jl
ngomezve: Done

src/cryo_tank/tanktools.jl
ngomezve: Right indeed

src/cryo_tank/tanktools.jl
ngomezve: I think the ODE solution should always converge. It is basically an integrator. Whether it converges to the right thing is a different question, but that's impossible to know unless you run another solver.

askprash:
Taking another look at this now before pulling it into main.

askprash:
@ngomezve the type stability issue you noticed is more wide spread with anything that touches the fuselage. I spent some time trying several options. This one seem to fix your issues (same speed on my mac with your head and slightly reduced allocations) but also prevents any future type stability issues with the cross-sections. It avoids having to create (and therefore track and maintain) several functions across the files just to get the fields of a struct.

There are several other places in the code that suffer from similar problems that I'll take a look at later. 

Take a look and let me know what you think!

src/IO/read_input.jl
```diff
@@ -173,7 +172,7 @@ parg[igWpaymax] = maxpay
 parg[igfreserve] = readmis("fuel_reserves")
 parg[igVne] = Speed(readmis("Vne"))
 parg[igNlift] = readmis("Nlift")
-fuselage.cabin.exit_limit = exitlimit
+# fuselage.cabin.exit_limit = exitlimit
```
askprash, line 175:
```suggestion
```

ngomezve:
Very good suggestion @askprash! I hadn't thought of using parametric typing but it looks like that's the way to go. I also get lower run times and allocations in my benchmarks so you definitely are onto something!

askprash (approved):

PR #189, opened 2025-09-24
Aeroperformance Sweeps and Plots

argonaut22:
Exercising the aero models to generate the aircraft's aerodynamic characteristics via `aeroperf_sweep()` and the `DragPolar()` plots (see below)

Also added aerodynamic modelling summary page to the docs, corrected some other spots, handled the CL = 0 case for stability calcs (previously causing NaNs), and made airfoil plotter (`plot_airf(ac)`) accessible to main namespace.

image

image


src/aero/aeroperf_sweep.jl
```diff
@@ -0,0 +1,134 @@
...
+    Sample usage:
+
+        ```julia
+        CL_vec = [0.2:0.05:0.8...]
+        results_nb = aeroperf_sweep(ac, CL_vec, print_results=true)
+        ```
+
+See also: [`TASOPT.DragPolar`](@ref), [`TASOPT.balance_aircraft!`](@ref), [`TASOPT.aerodynamics.aircraft_drag!`](@ref).
+"""
+function aeroperf_sweep(ac_orig, CL_vec; imission=1, ip=ipcruise1, rfuel=1, rpay=1, ξpay=0.5,
```
askprash, line 40: Instead of doing this just based of whatever `ipcruise1` conditions are, I would make this more flexible to take in CL, Mach (i.e., what you need from the mission point to `balance_aircraft!` and get `aircraft_drag!`) this will let us get CL vs CD curves for a range of flight Mach numbers (and therefore different operating conditions). 

I don't think that will take too many changes from what you have. Perhaps a function `aeroperf(ac, CL, Mach; rfuel, rpay, \xipay)` that sets `ac.para[iaCL, ipdummy] = CL; ac.para[iaMach, ipdummy]= Mach` and simply does `balance_aircraft!()` to trim and `aircraft_drag` to return the drag components. 
`aeroperf_sweep` then just iterates through a range of CLs/Machs that you provide. Not providing one of them can then default to take the value from ipcruise1.

src/balance/balance.jl
```diff
@@ -168,6 +167,14 @@ function balance_aircraft!(ac, imission, ip, rfuel, rpay, ξpay, opt_trim_var; L
       CMVf1 = parg[igCMVf1]
       CLMf0 = parg[igCLMf0]
 
+      #edge case: if CL = 0, the center of pressure is undefined, calc raises NaNs.
```
askprash, line 170: Curious, what's your use case when you end up with a CL=0 here calculation?

askprash:
@argonaut22 This is great! Left some ideas and suggestions here.

src/aero/aeroperf_sweep.jl
argonaut22: good thought. will implement soonish

argonaut22:
ex of updated plots
image

PR #192, opened 2025-10-14
Drag breakdown / evolution plots

argonaut22:
Breakdown by specified flight point via `plot_drag_breakdown(ac)`. Can show as drag coeff. or drag fraction.

image

image


askprash:
This is neat! @argonaut22 We can rebase once #189 is merged. One thought, can we add mission point names instead of just numbers -- "Rotate", "Cruise 1"...

PR #194, opened 2025-10-20
Adding cruise deviation in TASOPT

aditeyashukla:
Added four additional deviation mission points between cruise1 and cruise2

Added mission parameters imDeviationDH, imDeviationDL, imDeviationStart (distance after cruise1 where deviation should start)

Redid mission iteration cruise calculation

Work left to be done:
- Currently imDeviationDL is from first deviation point to last deviation point. It should actually be the distance between second and third deviation point (after climb).
-  Documentation and updated tests

### DIFF
--- a/src/mission/mission_iteration.jl
+++ b/src/mission/mission_iteration.jl
@@ -635,7 +635,6 @@ function _mission_iteration!(ac, imission, Ldebug; calculate_cruise = false)
                   # Limit climb γ to whichever is smaller: thrust capability or climb leg constraint
                   γ_cap = asin(sin_cap)
                   γ_eff = min(γ_abs, γ_cap)
-                  println("$(γ_cap), $(γ_eff)")
             end
       end
 
--- a/example/defaults/default_input.toml
+++ b/example/defaults/default_input.toml
@@ -112,8 +112,8 @@
 
 [Mission.Deviation]
     deviation_height = ["0.0 ft", "1000.0 ft"]
-    deviation_length = ["0.0 nmi", "50 nmi"]
-    deviation_start_from_TOC = ["0.0 nmi", "200.0 nmi"]
+    deviation_length = ["0.0 nmi", "100 nmi"]
+    deviation_start_from_TOC = ["0.0 nmi", "100.0 nmi"]
 
 [Mission.Descent]
     descent_angle_top-of-descent = "-3.0 deg"
--- a/src/IO/output_plots.jl
+++ b/src/IO/output_plots.jl
@@ -676,6 +676,33 @@ function plot_details(ac::aircraft)
 end
 
 """
+    plot_profile(ac::aircraft, imission::Integer)
+
+Plot altitude vs. range along with the trajectory climb angle for mission `imission`.
+"""
+function plot_profile(ac::aircraft, imission::Integer)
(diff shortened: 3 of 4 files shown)


src/mission/mission_iteration.jl
```diff
@@ -435,8 +435,6 @@ function _mission_iteration!(ac, imission, Ldebug; calculate_cruise = false)
             BW = W + para[iaWbuoy, ip]
             F = BW * (DoL + para[iagamV, ip])
             pare[ieFe, ip] = F / parg[igneng] #Store required thrust for engine calcs
-            Wpay = parg[igWpay]
-
```
askprash: What happened here?

aditeyashukla:
@askprash I think `mission_iteration.jl` works now  - the code is a bit long and messy but wanted to keep it as simple as possible, not trying to do too much in one place. 
After you have a look and see if what Im doing is right, I can clean up and also update other tests to be ready to push

src/data_structs/index.inc
```diff
@@ -503,14 +506,18 @@ ncrowx = ieTmet1 - ieepsc1
...
-      const ipcruise2 = 11
-      const ipdescent1 = 12
-      const ipdescent2 = 13
-      const ipdescent3 = 14
-      const ipdescent4 = 15
-      const ipdescent5 = 16
-      const iptest = 17
-      const iptotal = 17 
+      const ipdeviation1 = 11
+      const ipdeviation2 = 12
```
askprash: This is fine for now, but we really ought to implement the mission state/ segment/ maneuver approach to the mission "points"

src/mission/mission_iteration.jl
```diff
@@ -480,102 +478,385 @@ function _mission_iteration!(ac, imission, Ldebug; calculate_cruise = false)
...
+      S_dev_req = parm[imDeviationStart]
+
+      cruise_indices = (ipcruise1, ipdeviation1, ipdeviation2, ipdeviation3, ipdeviation4, ipcruise2)
+      ncr = length(cruise_indices)
+      # Track cumulative range from cruise start for each waypoint
+      s_values = zeros(Float64, ncr)
+      # Altitude offset relative to the nominal cruise-climb line
+      offsets = zeros(Float64, ncr)
+      # γ assigned to each cruise segment (piecewise constant between indices)
+      segment_gammas = fill(gamVcr1, ncr - 1)
```
askprash: Since you've already extended the mission points `ip...` you could use the existing `iaRange, iagamV, iaalt` etc, no?

src/mission/mission_iteration.jl
```diff
@@ -480,102 +478,385 @@ function _mission_iteration!(ac, imission, Ldebug; calculate_cruise = false)
...
+                                    use_deviation = false
+                              else
+                                    coef_total = (1.0 / den_up) - (1.0 / den_dn)
+                                    if abs(coef_total) < eps_den
+                                          if Ldebug
+                                                warn("Deviation skipped: deviation geometry becomes ill-conditioned (denominator nearly zero).")
+                                          end
+                                          use_deviation = false
+                                    else
+                                          Δh_limit = dR_available / abs(coef_total)
```
askprash: I'm not sure I follow what you are doing here.

PR #218, opened 2026-07-20
Modify the engine spanwise location setup in wing_weights!() function

ychen172:
**Changes proposed in this pull request**
- Change the y-engine location input to wing_weights! from etas to (b/2)*(etas-etao)
- Change docstring comment for y-engine inputs to wing_weights!

**Description**
- In wing_weights!, y engine location (dyeinn) is used to compute the engine weight induced wing-root bending moment relief (https://github.com/MIT-LAE/TASOPT.jl/blob/b9be1deb59879d5be3b3f4d9a964d173a4fc20e0/src/structures/wing_weights.jl#L146). 
- If comparing with the definition of other terms like wing.inboard.dyW, I think the dyeinn in the last term should be dimensional (m) and measured outward positive from wing root. 
- However,  from the function call inside size_aircraft (https://github.com/MIT-LAE/TASOPT.jl/blob/b9be1deb59879d5be3b3f4d9a964d173a4fc20e0/src/sizing/size_aircraft.jl#L365), a normalized parameter eta_s is provided, which is between 0->1 (or interpreted as 0->1 m I believe). 
- Thus, I think that input should be replaced with dimensional distance like "0.5*b*(eta_s-eta_o)" instead.
- Also the description in docstring about both inboard and outboard y-engine being measured from wingspan is modified

**Example**
- I was comparing the optimization results for etas before and after this modification
PFEI
eta_s

- I noticed the optimal span break location (40%) was higher than reference aircraft like B737 before the modification. I thought this was caused by the used of non-dimensional distance that the optimizer is trying to increase etas for more bending relief as if etas were dimensional.
- However, from the optimization result, it seems that after modification, optimizer would push etas to even a higher value (60%), while this push did improve the PFEI from the case before. So maybe this is the bending relief really taking effect.
- Based on the results, I have a conflict feeling towards whether this is a real fix or not.

 **Thought for next step**
-  My original intention or a further extension beyond this setup is to input the dyeinn from size_aircraft instead as "yeng - 0.5*wing.layout.span*wing.layout.ηo". The reason is yeng is actually a input parameter used to size the vertical tail size if I turn on the OEI vtail sizing option (Sizing tail area based on one engine out condition during rotation).
- My intention is that if yeng were too far out, then vertical tail will become too large and penalize PFEI. Simultaneously, if yeng were too short, wing root bending relief will be insufficient and penalize PFEI. I was hoping this tradeoff will truly create a optimal point for yeng instead of single sidedly very out and very in. And this will also isolated out yeng position from span-break position, whose optimal will come from instead wing geometric and aero sizing. 
- Unfortunately, during the optimization, it was found out that when OEI option for vtails sizing were turned on, then yeng will be optimized to be as little as possible. Basically, the penalty of increasing vertical tail size outweights that of the bend stress relief very easily. (If turn off OEI, yeng becomes as far as possible as expected)
- As this vertical tail sizing coupling is something I am still currently looking into, I think the current change I wanna propose is simply etas-etao, assuming engine is at the etas location.

**AI Statement (required)**

- **Limited use of generative AI.**
  Standard or boilerplate code snippets were generated with AI and manually reviewed;
  all design, logic, and implementation decisions were made by the contributor.
  Examples: IDE code-completions or brief LLM queries for common patterns.

 -->

**Checklist**

- [ ] The pull request includes a clear description of this code change
- [ ] Commit messages have short titles and reference relevant issues
- [ ] All tests are passing
- [ ] AI Statement is included
- [ ] The pull request is ready for review


ychen172:
Let me include this in a more complete pull request right after this.

